PR #11, opened 2021-06-07
add solution

iedenysenko:


src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+        throw new AuthenticationException("Can't authenticate user by email: " + email);
+    }
+    
+    @Override
+    public User register(String email, String password) {
+        User user = new User();
+        user.setLogin(email);
+        user.setPassword(password);
+        return userService.add(user);
+    }
```
SerhiiKysliakov, line 34: what will be if user already exists?

SerhiiKysliakov (changes requested):
i don't see hibernate.cfg.xml file (how do you connect to DB?)
where is you main method? you should test your implementation!

Resci (approved):

PR #15, opened 2021-06-08
Added User, UserService, UserDao classes and implementation

Den4eg007:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", salt=" + Arrays.toString(salt)
```
dvdovynets: Do we need to add salt in toString?

dvdovynets (approved):

Resci (approved):

PR #68, opened 2021-08-09
AL added User, UserService, UserDao, Authentication

aaananas:


src/main/java/mate/academy/Main.java
```diff
@@ -1,55 +1,55 @@
...
-        System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
-        System.out.println(movieSessionService.findAvailableSessions(
-                        fastAndFurious.getId(), LocalDate.now()));
+    public static void main(String[] args) {
+        AuthenticationService authenticationService = (AuthenticationService)
+                injector.getInstance(AuthenticationService.class);
+        UserService userService = (UserService) injector.getInstance(UserService.class);
+        User bob = new User();
+        bob.setName("Bob");
+        bob = authenticationService.register("bob@mail.com", "bob123");
```
Basilius-Third, line 18: Does your user even have name here? I think this operation overwrites your user link with another users (the one without a name) link. Same for else

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+                return user;
+            }
+        }
+        throw new AuthenticationException("Can't authenticate user email: " + email);
+    }
+
+    @Override
+    public User register(String email, String password) {
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RuntimeException("This email " + email + " is already registered");
```
Basilius-Third, line 33: It is better to throw custom RegistrationException and it's also shouldn't extend RuntimeException

Basilius-Third (changes requested):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
aaananas: 👍

src/main/java/mate/academy/Main.java
aaananas: 👍

Basilius-Third (approved):

Sofasmile (approved):

PR #72, opened 2021-08-11
Hibernate-user-service-hw-solution

olekskov:
Initial solution:
- Created User model/dao/service levels
- Created AuthenticationService
- Created HashUtil for password hashing with SHA-512 using salt
- Tested new functionality in PSVM
- MovieSessionDaoImpl corrected just to make PSVM existing methods work


RomanDubovskyi (approved):
all good

PR #75, opened 2021-09-13
added solution to hw hibernate user service

mtv1301:


src/main/java/mate/academy/security/AuthenticationService.java
```diff
@@ -0,0 +1,12 @@
+package mate.academy.security;
+
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.model.User;
+
+public interface AuthenticationService {
+
+    User login(String email, String password) throws AuthenticationException;
```
boroda4436, line 9: Could you please verify this method works well? Can we add a corresponding example to the `main` method?

boroda4436 (approved):

PR #87, opened 2021-10-27
hibernate-user-service

dogovor:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,31 @@
...
+            }
+            return hashedPwd.toString();
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Can`t hash password!", e);
+        }
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom random = new SecureRandom();
+        byte[] salt = new byte[16];
```
Zbarazkyy, line 27: Size of the array can be a constant

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
Zbarazkyy, line 15: email must be unique, write annotation over field

Zbarazkyy (changes requested):
Everything is good, but you forgot to put annotation over the field

Zbarazkyy (approved):

Resci (approved):

PR #104, opened 2021-10-31
User service

DenysST:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Could not crate hash using SHA-512 algorithm",e);
```
netiazhevskyi, line 31: you have type in word create:)

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDB = userService.findByEmail(email);
```
netiazhevskyi, line 20: you can throw exception first if user is empty and then do the rest
Or you can do something like this
        if (user.isEmpty() || compare passwords)

netiazhevskyi (approved):

Resci (approved):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
Resci: Agree

PR #114, opened 2021-11-01
implemented user service and user dao

ditroRV:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByLogin(email).isEmpty()) {
+            return userService.save(new User(email, password));
+        }
+        throw new RegistrationException("User with email" + email + "already exists");
+    }
+
+    @Override
+    public User login(String login, String password) {
```
VitaliiMYKH, line 27: you can remake that method according to checklist, by using only one if statement

VitaliiMYKH (approved):

Resci (approved):

PR #122, opened 2021-11-08
Hibernate user service solution

lena-kravchenko:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,36 @@
+package mate.academy.model;
+
+public class User {
```
MaliukDaria, line 3: We need to make `User` as entity class to work with it through Hibernate

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,36 @@
...
+        this.password = password;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
```
MaliukDaria, line 35: we need to define `toString()` method to log user entity in exception messages

src/main/java/mate/academy/service/impl/AuthentificationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can't authentificate user.");
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Email or password was incorrect.");
```
MaliukDaria, line 29: lets use only one `if` here (and one exception message)

src/main/java/mate/academy/service/impl/AuthentificationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        User user = new User(email, password);
+        try {
+            return userService.add(user);
+        } catch (Exception e) {
+            throw new RegistrationException("Can't register user.");
+        }
```
MaliukDaria, line 39: here we need to check if user with such email already exists in the db, if yes - throw RegistrationException, if not - register

MaliukDaria (changes requested):

src/main/java/mate/academy/model/User.java
lena-kravchenko: fixed

src/main/java/mate/academy/model/User.java
lena-kravchenko: done

src/main/java/mate/academy/service/impl/AuthentificationServiceImpl.java
lena-kravchenko: fixed

src/main/java/mate/academy/service/impl/AuthentificationServiceImpl.java
lena-kravchenko: fixed

MaliukDaria (approved):
Good Job 🔥

PR #123, opened 2021-11-16
1.0V Solution "user-service"

Chupiak:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        User user = userService.findByEmail(email).get();
```
RomanDubovskyi, line 19: if user is not present you will get here NoSuchElementException instead of AuthenticationException
Let's check if user is present if passwords are the same in other cases throw AuthenticationException

RomanDubovskyi (changes requested):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -16,11 +16,16 @@
 
     @Override
     public User login(String email, String password) throws AuthenticationException {
+        if (userService.findByEmail(email).isEmpty()) {
```
RomanDubovskyi, line 19: Ideally, make single if clause to shorten the code

RomanDubovskyi (approved):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
Chupiak: Ok, i will do it

PR #165, opened 2022-01-01
add UserDao, UserService, AuthenticationService

annaRav:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
```
MaliukDaria, line 12: lets add toString method(without salt and password fields) to display user correct in exception messages

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import mate.academy.service.UserService;
+
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> byEmail = userService.findByEmail(email);
+        if (byEmail.isEmpty() || !byEmail.get().getPassword().equals(password)) {
```
MaliukDaria, line 18: We have hashed password in db and unhashed password that comes to the input parameter to the method, they could not be equals(Lets hash password first)

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+            throw new AuthenticationException("Can't find User with email = "
+                + email + " and password = " + password);
```
MaliukDaria, line 20:
```suggestion
            throw new AuthenticationException("Can't find User with email = "
                + email);
```

lets remove sensitive data from exception message

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        return byEmail.get();
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        return userService.add(new User(email, password));
+    }
```
MaliukDaria, line 29: where can we get RegistrationException here?

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,31 @@
...
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Can't create hash using "
+                + CRIPTO_ALGORITM + " algorithm", e);
+        }
+        return handlePassword.toString();
+    }
+
+    public static byte[] getSalt() {
+        return new byte[]{12}; //todo
```
MaliukDaria, line 29: lets use SecureRandom class here

MaliukDaria (changes requested):
What about hibernate.cfg.xml file? Do you test your solution, does it works without it?

MaliukDaria (approved):

PR #168, opened 2022-01-19
Wrote an assignment hibernate-user-service-hw.

ASelivonchyk:
- configured hibernate.cfg.xml
- added POJO User;
- created UserDao and it Impl with methods add(), findByEmail();
- created UserService and it Impl. Method findByEmail() in UserServiceImpl convert string pass to hash pass;
- added HashUtil class to get salt and hash passwords;
- added Authentication & Registration exceptions;
- created AuthenticationService and its impl with methods login(), register();
- in Main class tested workability.


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
+    @Column(unique = true)
```
MaliukDaria, line 17: Why do we need password to be unique?

MaliukDaria (changes requested):

src/main/java/mate/academy/model/User.java
ASelivonchyk: Yea, it`s a mistake, annotation shuld be above email.

MaliukDaria (approved):

PR #173, opened 2022-03-02
Home work

Noct2000:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,62 @@
...
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt) + '}';
```
MaliukDaria, line 60: password and salt it's sensitive data, lets remove it from toString method

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import mate.academy.lib.Inject;
+import mate.academy.lib.Service;
+import mate.academy.model.User;
+import mate.academy.service.AuthenticationService;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final String ERROR_AUTH_MES = "Incorrect email or password";
```
MaliukDaria, line 15: I dont think we need to make exception message as constant

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #179, opened 2022-03-24
Authentication with password encryption was implemented for User model

karma-o:


src/main/java/mate/academy/Main.java
```diff
@@ -11,7 +12,9 @@
 
 public class Main {
     public static void main(String[] args) {
-        MovieService movieService = null;
```
pikador10, line 14: I don't see register user and login some user in method main

src/main/java/mate/academy/exception/RegistrationException.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.exception;
+
+public class RegistrationException extends Exception {
+    public RegistrationException(String message) {
+        super(message);
+    }
+
+    public RegistrationException(String message, Throwable cause) {
+        super(message, cause);
```
pikador10, line 9: You can remove cause, because you don't use it

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-256";
+
+    public HashUtil() {
+    }
```
pikador10, line 11: Look checklist -> Let’s add private constructor to HashUtil.

pikador10 (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +52,24 @@ public static void main(String[] args) {
...
+
+        UserService userService = (UserService) injector.getInstance(UserService.class);
+        AuthenticationService authenticationService =
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        try {
+            userService.add(authenticationService.register("alice@gmail.com", "1234"));
+            userService.add(authenticationService.register("bob@gmail.com", "qwerty"));
+            System.out.println(authenticationService.login("alice@gmail.com", "1234"));
+        } catch (RegistrationException | AuthenticationException e) {
+            e.printStackTrace();
```
boroda4436, line 72: This is a bad practice. If you are catching exception in place like main method - it's better to print out meaningful message

boroda4436 (approved):

PR #188, opened 2022-04-08
hibernate user service implemented

sltnsrh:


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+                return user;
+            }
+        }
+        throw new AuthenticationException("There is no user with this email " + email
+        + " or password is incorrect");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        User user = null;
```
pikador10, line 33: In this method also need to check if user with this email is present in DB, if present throw Ex

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+import mate.academy.service.UserService;
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Inject
+    private UserDao userDao;
+
+    @Override
+    public User add(User user) {
+        return userDao.add(user);
```
pikador10, line 17: If we save user, what kind of password was added to DB, with hash or without hash

pikador10 (approved):

KostyaNikiforov (approved):

PR #191, opened 2022-04-11
all needed classes were implemented

DzhuzhaMichael:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +50,27 @@ public static void main(String[] args) {
...
+
+        User secondUser = new User();
+        secondUser.setEmail("second@mail.com");
+        secondUser.setPassword("secondpa$$word");
+
+        UserService userService = (UserService) injector.getInstance(UserService.class);
+        userService.add(firstUser);
+        userService.add(secondUser);
+        System.out.println(userService.findByEmail(firstUser.getEmail()));
+        System.out.println(userService.findByEmail(secondUser.getEmail()));
```
o240789, line 74: I think we need to test AuthenticationService here (register() and login())

o240789 (approved):

src/main/java/mate/academy/Main.java
DzhuzhaMichael: Many thanks for your comment! The main method was fixed!

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can`t authenticate user with email " + email);
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (email.isEmpty() || password.isEmpty()) {
+            throw new RegistrationException("Email or password should not be empty");
```
KostyaNikiforov, line 35:
```suggestion
            throw new RegistrationException("Email and password should not be empty");

```

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isPresent()) {
+            throw new RegistrationException("User with email " + email + " already exists");
+        }
```
KostyaNikiforov, line 40:
```suggestion
        if (userService.findByEmail(email).isPresent()) {
            throw new RegistrationException("User with email " + email + " already exists");
        }

```
I think there is no need to create a named object if we only use it once.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    private HashUtil() {
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] salt = new byte[16];
```
KostyaNikiforov, line 15: Let's make this value constant too

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can`t authenticate user");
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can`t authenticate user with email " + email);
+    }
```
KostyaNikiforov, line 30: We are throwing almost the same exceptions twice. Can we combine it in one?

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #192, opened 2022-04-11
Hibernate User Service - Initial commit

o240789:


DzhuzhaMichael (approved):
Looks good to me!

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("User email or password is incorrect");
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        } else {
+            throw new AuthenticationException("User email or password is incorrect");
+        }
```
KostyaNikiforov, line 29: Let's refactor this to throw the same exception just once. Or you can put the message in a constant

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userByEmail = userService.findByEmail(email);
+        if (userByEmail.isPresent()) {
+            throw new RegistrationException("User with this email is already registered");
+        }
```
KostyaNikiforov, line 37:
```suggestion
        if ( userService.findByEmail(email).isPresent()) {
            throw new RegistrationException("User with this email is already registered");
        }
```
I think there is no need to create a named object if we only use it once.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    private HashUtil() {
+
```
KostyaNikiforov, line 11: Empty line

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] salt = new byte[16];
```
KostyaNikiforov, line 16: Let's make this value constant too

KostyaNikiforov (changes requested):

PR #195, opened 2022-04-12
added model User, added Dao and service layer for this model. added a…

black-bit0:
…uthentication service

### DIFF
--- a/src/main/java/mate/academy/security/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/academy/security/AuthenticationServiceImpl.java
@@ -1,0 +1,42 @@
+package mate.academy.security;
+
+import java.util.Optional;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Inject;
+import mate.academy.lib.Service;
+import mate.academy.model.User;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User register(String login, String password) throws RegistrationException {
+        Optional<User> userFromDB = userService.findByLogin(login);
+        if (userFromDB.isPresent()) {
+            throw new RegistrationException("Login " + login + " already in use");
+        }
+        User user = new User();
+        user.setLogin(login);
+        user.setPassword(password);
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User register(String login, String password) throws RegistrationException {
+        Optional<User> userFromDB = userService.findByLogin(login);
+        if (userFromDB.isPresent()) {
+            throw new RegistrationException("Login " + login + " already in use");
+        }
```
KostyaNikiforov, line 21: There is no need to create a named object if we use it only once
```suggestion
        if (userService.findByLogin(login).isPresent()) {
            throw new RegistrationException("Login " + login + " already in use");
        }

```

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+        if (userFromDB.isEmpty()) {
+            throw new ArithmeticException("Can`t find user where login is " + login);
+        }
+        User user = userFromDB.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new ArithmeticException("Can`t find user where login is " + login);
+    }
```
KostyaNikiforov, line 41: You are throwing the same exception twice. Can we combine this into one

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    private HashUtil() {
+
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom random = new SecureRandom();
+        byte[] salt = new byte[16];
```
KostyaNikiforov, line 16: Let's make this value constant too

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #254, opened 2022-07-13
implement task

gdpbalt:
- add entity User;
- add dao for working with User;
- add service for working with User;
- add exceptions RegistrationException and AuthenticationException;
- add AuthenticationService with methods login() and register();
- add class HashUtil for hashing plain password;
- add usage examples to Main class.

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,54 +2,82 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Injector;
 import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.User;
+import mate.academy.security.AuthenticationService;
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.UserService;
 
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.academy");
+    private static final MovieService movieService =
+            (MovieService) injector.getInstance(MovieService.class);
+    private static final CinemaHallService cinemaHallService =
+            (CinemaHallService) injector.getInstance(CinemaHallService.class);
+    private static final MovieSessionService movieSessionService =
+            (MovieSessionService) injector.getInstance(MovieSessionService.class);
+    private static final UserService userService =
+            (UserService) injector.getInstance(UserService.class);
+    private static final AuthenticationService authenticationService =
+            (AuthenticationService) injector.getInstance(AuthenticationService.class);
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        try {
+            String email = "jack@example.com";
+            String password = "12345";
+            authenticationService.register(email, password);
+            User jack = authenticationService.login(email, password);
+            System.out.println("* ".repeat(10));
+            System.out.println(jack);
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Registration error", e);
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("Authentication error", e);
+        }
+        System.out.println("* ".repeat(10));
+        userService.getAll().forEach(System.out::println);
 
         Movie fastAndFurious = new Movie("Fast and Furious");
         fastAndFurious.setDescription("An action film about street racing, heists, and spies.");
         movieService.add(fastAndFurious);
-        System.out.println(movieService.get(fastAndFurious.getId()));
+
+        System.out.println("* ".repeat(10));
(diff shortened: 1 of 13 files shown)


src/main/java/mate/academy/security/AuthenticationService.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface AuthenticationService {
+    User login(String email, String password) throws AuthenticationException;
+
+    /**
+     * We should register a new user. New user entity will contain the email and password
+     * @param email - user email. should be unique for each user
+     * @param password - user password
+     * @return new user instance
+     */
```
antonEEmus, line 15: remove this comment

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+            if (user.getPassword().equals(hashedPassword)) {
+                return user;
+            }
+        }
+        throw new AuthenticationException("Email or password are wrong");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByEmail(email).isPresent()) {
```
antonEEmus, line 32: add a check for an empty password here

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-256";
+    private static final int SALT_LONG = 16;
```
antonEEmus, line 9:
```suggestion
    private static final int SALT_LENGTH = 16;
```

antonEEmus (changes requested):

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
gdpbalt: I don't agree
Because verification of input data it's anothe service, for example AuthenticationFormVerificationService
Also password can be verified minimum and maximum long

src/main/java/mate/academy/Main.java
```diff
@@ -2,54 +2,82 @@
...
+    private static final MovieService movieService =
+            (MovieService) injector.getInstance(MovieService.class);
+    private static final CinemaHallService cinemaHallService =
+            (CinemaHallService) injector.getInstance(CinemaHallService.class);
+    private static final MovieSessionService movieSessionService =
+            (MovieSessionService) injector.getInstance(MovieSessionService.class);
+    private static final UserService userService =
+            (UserService) injector.getInstance(UserService.class);
+    private static final AuthenticationService authenticationService =
+            (AuthenticationService) injector.getInstance(AuthenticationService.class);
```
KostyaNikiforov, line 29: I think, there is no reason to make service classes static. Anyway, the main class is created only once and imho it will look better if they are class fields.

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            CriteriaBuilder criteriaBuilder = session.getCriteriaBuilder();
+            CriteriaQuery<User> criteriaQuery = criteriaBuilder.createQuery(User.class);
+            Root<User> userRoot = criteriaQuery.from(User.class);
+            criteriaQuery.where(criteriaBuilder.equal(userRoot.get("email"), email));
+            return session.createQuery(criteriaQuery).uniqueResultOptional();
```
KostyaNikiforov, line 68: As practice is it cool! But next time it will be easier to use `createQuery` method. We use createriaQuery only in case we don't know how many params we will get.

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get all users", e);
```
KostyaNikiforov, line 70:
```suggestion
            throw new DataProcessingException("Can't find user by email: " + email, e);
```

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public User add(User user) {
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
+        return userDao.add(user);
+    }
+
+    @Override
+    public User get(Long id) {
+        return userDao.get(id).get();
```
KostyaNikiforov, line 26:
```suggestion
        return userDao.get(id).orElseThrow();
```
It is just a bit clearer

KostyaNikiforov (changes requested):

PR #255, opened 2022-07-19
created new classes

Anastasia-702:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("User with this email already exists " + email);
+        }
+        if (password.isEmpty()) {
+            throw new RegistrationException("Invalid password");
+        }
```
Nick97-git, line 35: let's remove this check, in the future you will be making check of password in a different way

Nick97-git (changes requested):

Nick97-git (approved):

PR #256, opened 2022-07-19
Chumak user hibernate service

ChumakAndrey:


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (email.isEmpty()) {
+            throw new RegistrationException("Wrong email");
+        }
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("User with such email exist already");
+        }
```
gdpbalt, line 37: In this place you check email
1. Why don't you check password too ?
2. Can you join both if expression in one (I think mentor will want this too) ?
3. Why don't you check password in method login ?

gdpbalt (changes requested):

gdpbalt (approved):

liliia-ponomarenko (approved):
LGTM!

PR #261, opened 2022-07-23
task done

nazarbaklanov:
added model User
created class UserDao
created UserService and AuthenticationService
added own checked RegistrationException and AuthenticationException

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,16 +2,40 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Injector;
 import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.User;
+import mate.academy.service.AuthenticationService;
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
 
 public class Main {
+    private static final Injector injector
+            = Injector.getInstance("mate.academy");
+    private static final MovieService movieService
+            = (MovieService) injector.getInstance(MovieService.class);
+    private static final MovieSessionService movieSessionService
+            = (MovieSessionService) injector.getInstance(MovieSessionService.class);
+    private static final CinemaHallService cinemaHallService
+            = (CinemaHallService) injector.getInstance(CinemaHallService.class);
+    private static final AuthenticationService authenticationService
+            = (AuthenticationService) injector.getInstance(AuthenticationService.class);
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        try {
+            User alice = authenticationService.register("test@gmail.com", "qwerty");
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,64 @@
...
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", name='" + email + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt)
```
Cosmic-M, line 61: last cause in checkStyle file! You should remove it

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<User> query = session.createQuery("FROM User WHERE email = :email", User.class);
+            query.setParameter("email", email);
+            return query.uniqueResultOptional();
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get user by email: " + email, e);
+        }
+
```
Cosmic-M, line 46: empty line

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,16 @@
+<hibernate-configuration>
+    <session-factory>
+        <property name="hibernate.dialect">org.hibernate.dialect.MySQL5Dialect</property>
+        <property name="connection.url">jdbc:mysql://localhost:3306/cinema_app?serverTimezone=UTC</property>
+        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>
+        <property name="connection.username">root</property>
+        <property name="connection.password">bna123456</property>
+        <property name="show_sql">true</property>
+        <property name="hibernate.hbm2ddl.auto">create-drop</property>
```
Cosmic-M, line 9: if you apply update option, you'll be able to test your try-catch construction in main. Now your construction (try-catch in maim, I mean) has no matter, because you don't test option if you add user with email which exists in DB.

src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,40 @@
...
+            = (MovieSessionService) injector.getInstance(MovieSessionService.class);
+    private static final CinemaHallService cinemaHallService
+            = (CinemaHallService) injector.getInstance(CinemaHallService.class);
+    private static final AuthenticationService authenticationService
+            = (AuthenticationService) injector.getInstance(AuthenticationService.class);
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        try {
+            User alice = authenticationService.register("test@gmail.com", "qwerty");
```
Cosmic-M, line 31: you assign variable alice which you don't use

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
Cosmic-M: I just thinking about, we both have the same imprecision in block catch: I think, it will be better to do message like this:
"Can't find user by email=" + email + " from DB or data isn't unique", e);

Cosmic-M (approved):

KostyaNikiforov (approved):

PR #306, opened 2022-10-02
user registration and login implemented

max-bolotin:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,60 @@
...
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
+    private String password;
+    private byte[] salt;
+
+    public User() {
+    }
```
GolDenik, line 20: It is redundant . You use it by default

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+                System.out.println("User with email=" + email + " logged in successfully!");
+                return user;
+            }
+        }
+        throw new AuthenticationException("Incorrect password or login ");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
```
GolDenik, line 36: You don't need session here. And you can't work with DB on Service layer.

GolDenik (changes requested):

GolDenik (approved):

src/main/java/mate/academy/dao/UserDao.java
```diff
@@ -0,0 +1,12 @@
+package mate.academy.dao;
+
+import java.util.Optional;
+import mate.academy.lib.Dao;
+import mate.academy.model.User;
+
+@Dao
```
Sofasmile, line 7: should we use Dao annotation above the interface?

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        try {
+            session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            session.save(user);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new RuntimeException("Can't save user in DB", e);
```
Sofasmile, line 29: which user? please add more information to the message

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isPresent()) {
+            User user = userFromDbOptional.get();
+            String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+            if (user.getPassword().equals(hashedPassword)) {
+                System.out.println("User with email=" + email + " logged in successfully!");
```
Sofasmile, line 25: we shouldn't use sout, please remove it

Sofasmile (changes requested):

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -26,7 +26,8 @@ public User add(User user) {
             if (transaction != null) {
                 transaction.rollback();
             }
-            throw new RuntimeException("Can't save user in DB", e);
+            throw new RuntimeException("Can't save user with email="
+                    + user.getEmail() + " in DB", e);
```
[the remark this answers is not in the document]
max-bolotin: Why do we have to receive excessive info about user in this message? I believe just user's login (an email in this case) would be enough. I'll correct this but I'm not sure if it's really necessary

Sofasmile (approved):

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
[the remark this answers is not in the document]
Sofasmile: sometimes it is essential to know all info about an object which can't be saved to DB. Because the reason can be in the content of some field. But you don't know what kind of field it might be.

PR #325, opened 2022-10-04
done User service hw

Vasul-Pankiv:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Arrays;
+import java.util.Objects;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+@Entity
```
valentyn-slutskyi, line 11: You can add a name of the table, like "users"

valentyn-slutskyi (changes requested):
I like it)

src/main/java/mate/academy/dao/UserDao.java
```diff
@@ -0,0 +1,12 @@
+package mate.academy.dao;
+
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserDao {
```
valentyn-slutskyi, line 6:
```suggestion
@Dao
public interface UserDao {
```
It is necessary for injector

valentyn-slutskyi (changes requested):

src/main/java/mate/academy/dao/UserDao.java
Vasul-Pankiv: Yes? but we need this annotation only in daoImpl

valentyn-slutskyi: I'm so sorry) Yes, it is my mistake)

valentyn-slutskyi (approved):

Noct2000 (approved):

PR #368, opened 2022-11-28
Hibernate user service hw

SoulARC:


shmex-j (approved):
Good job!

src/main/java/mate/academy/model/CinemaHall.java
```diff
@@ -5,46 +5,18 @@
...
 
 @Entity
 @Table(name = "cinema_halls")
+@Getter
+@Setter
+@ToString
 public class CinemaHall {
-    @Id
-    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
```
ChumakAn, line 18: it's better to write 1 annotation per line

ChumakAn: same for all

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.service;
+
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.model.User;
+
+public interface AuthenticationService {
+    User login(String email, String password) throws AuthenticationException, RegistrationException;
+
+    User register(String email, String password) throws AuthenticationException;
```
ChumakAn, line 10: why method login throws RegistrationException? and the the register method does not? Maybe you need to change either logic or namings

src/main/java/mate/academy/service/impl/CinemaHallServiceImpl.java
```diff
@@ -19,7 +20,8 @@ public CinemaHall add(CinemaHall cinemaHall) {
 
     @Override
     public CinemaHall get(Long id) {
-        return cinemaHallDao.get(id).get();
+        return cinemaHallDao.get(id).orElseThrow(() ->
+        new NoSuchElementException("No cinema hall by id:" + id));
```
ChumakAn, line 24: what about the EntityNotFoundException?

ChumakAn: for all

ChumakAn (changes requested):
good! Let's make it better 🔥

ChumakAn (approved):
LGTM

PR #382, opened 2022-12-07
solution for user-service

Bohdan20b:


anton-pushkar (approved):

anton-pushkar:
Good Job 👌

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public Optional<User> get(Long id) {
+        return userDao.get(id);
+    }
+
+    @Override
+    public List<User> getAll() {
+        return userDao.getAll();
+    }
```
ChumakAn, line 32: it's not obligatory to add this methods

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public User add(User user) {
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(),user.getSalt()));
+        return userDao.add(user);
+    }
+
+    @Override
+    public Optional<User> get(Long id) {
+        return userDao.get(id);
```
ChumakAn, line 26: but it will be a good strategy to return User object in the service layer, and use orElseThrow method, throwing EntityNotFoundException with info message

ChumakAn (approved):
Good! Take my comments into account

PR #384, opened 2022-12-07
added user service and authentication service to app

apc10bbe:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,85 @@
...
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @Column(unique = true)
+    private String login;
+    private String password;
+    @Transient
+    private String repeatPassword;
```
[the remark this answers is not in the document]
ChumakAn: I'm not sure, you need this, usually, we use this field on front-end part to validate if the user entered right password, but we don't save it to db for example and don't need this as a field

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            throw new RegistrationException("Your passwords not equals. Please try again.");
+        }
+        return userService.add(new User(login, password));
+    }
+
+    private boolean isPasswordValid(User user, String password) {
+        return user.getPassword().equals(HashUtil.hashPassword(password, user.getSalt()));
+    }
+
+    private boolean isUserExist(String login) {
```
ChumakAn, line 45:
```suggestion
    private boolean isLoginValid(String login) {
```

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public static String hashPassword(String password, byte[] salt) {
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
```
ChumakAn, line 28: it can be a constant

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
...
+            }
+
```
ChumakAn, line 30: empty line

ChumakAn (changes requested):
Good! Let's make it better

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
apc10bbe: year, repeat password not included to task. I just want to try do it. And I dont add this field to DB (using @Transient)

ChumakAn (approved):

PR #390, opened 2022-12-08
Hw hibernate user service

Mar4uk21:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,69 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
```
muzykaVP, line 66: [Don’t add password and salt to the toString() method.](https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-user-service/add-user-service-checklist)

muzykaVP (approved):
Looks good 👍 one minor common mistake

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDb = userService.findByEmail(email);
+        if (userFromDb.isPresent()) {
+            User user = userFromDb.get();
+            String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+            if (user.getPassword().equals(hashedPassword)) {
+                return user;
```
liliia-ponomarenko, line 25: better to avoid nesting of conditionals. Can you combine these conditionals?

liliia-ponomarenko (changes requested):
Good job! Let’s  fix a few mistakes ;)

src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,23 @@
...
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
 
 public class Main {
-    public static void main(String[] args) {
-        MovieService movieService = null;
+    private static final Injector injector = Injector.getInstance("mate.academy");
+
+    public static void main(String[] args) throws RegistrationException, AuthenticationException {
```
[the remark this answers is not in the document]
Mar4uk21: I can surround it with try-catch here, and when we will have a website we can redirect it to some endpoint.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDb = userService.findByEmail(email);
+        User user = userFromDb.get();
```
hronograf, line 21: what if we don't have user with such email in db?

src/main/java/mate/academy/Main.java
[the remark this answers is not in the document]
hronograf: let's add try-catch, as was suggested

hronograf (changes requested):

hronograf (approved):

PR #415, opened 2023-01-11
Hw31 hibernate user service hw

SichniyAndriy:


src/main/java/mate/academy/exception/RegistrationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class RegistrationException extends Exception {
+    public RegistrationException(String message, Throwable cause) {
```
aaananas, line 4:
```suggestion
    public RegistrationException(String message) {
```

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        User user = new User();
+        user.setEmail(email);
+        user.setPassword(password);
+        try {
+            user = userService.add(user);
+            return user;
+        } catch (NullPointerException e) {
```
aaananas, line 40: you should not catch NPE, it is not an exception related to the registration process. Something is not ok with your code if you get NPE

aaananas (changes requested):

aaananas (approved):

PR #424, opened 2023-02-03
build success, user methods implemented

Uka0001:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +53,23 @@ public static void main(String[] args) {
...
+
+        UserService userService = (UserService) injector.getInstance(UserService.class);
+        User userOne = new User("emailOne", "passwordOne");
+        userService.add(userOne);
+        System.out.println(userService.findByEmail(userOne.getLogin()));
+        AuthenticationService authenticationService
+                = (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        authenticationService.register("emailTwo", "passwordTwo");
+        authenticationService.register("emailNew", "passwordNew");
+        authenticationService.login("emailOne", "passwordOne");
```
Dmytro-Mk, line 73:
```suggestion
        authenticationService.register("emailTwo", "passwordTwo").sout;
        authenticationService.register("emailNew", "passwordNew").sout;
        authenticationService.register("emailNew", "passwordNew").sout;
        authenticationService.login("emailOne", "passwordOne").sout;
        authenticationService.login("emailOne", "wrongpassword").sout;
        authenticationService.login("wrongemail", "passwordOne").sout;
```
for proper testing )

Dmytro-Mk (changes requested):

src/main/java/mate/academy/Main.java
Uka0001: I don`t understand why we need to add try-catch in main if we have  already try catch in proper Dao an Controllers class?

Uka0001: and with sout we only see the user entity...  I think the same almost as just like default query messages of the hibernate ...

Dmytro-Mk (approved):

KostyaNikiforov (approved):

PR #456, opened 2023-02-27
All tasks are done.

svitlana-tertyshna:
Created model User, UserDao, UserService, AuthenticationService; added custom checked exceptions; tasted all methods in Main class.


kozhukhovsky (approved):
Good job!

PR #469, opened 2023-03-10
done

diana-lomei:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userOptional = userService.findByEmail(email);
+        if (userOptional.isPresent()) {
+            User user = userOptional.get();
+            String hashPassword = HashUtil.hashPassword(password, user.getSalt());
+            if (user.getPassword().equals(hashPassword)) {
+                return user;
+            }
+        }
+        throw new AuthenticationException("Login or password are incorrect");
```
okuzan, line 28: there should be one if according to the checklist :) 
image

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,62 @@
...
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt)
```
okuzan, line 59: don't help them hackers
image

okuzan (changes requested):

olekskov (approved):
Good job!

PR #470, opened 2023-03-10
Adds User and corresponding Dao and Service,

BusyhinD:
also implemented Authentication and Registration of the user.


Ivan95kos (approved):
Great job

PR #472, opened 2023-03-13
hibernate-user-service-solution

nkharyshyn:
-created model User with its DAO and Service layers; 
-created AuthenticationService with register and login methods; 
-created custom checked exceptions RegistrationException and AuthenticationException and thrown them in AuthenticationService;
-created HashUtil class with getSalt (to generate unique salt for each user) and hashPassword (to hash password);
-created configuration file hibernate.cfg.xml

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,62 @@
+package mate.academy.model;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
aaananas, line 15: see common mistakes

aaananas (changes requested):

okuzan (approved):
Good job

PR #475, opened 2023-03-15
Added user class, dao and service layers. Authentication service laye…

oAlExAnDeRo:
…rs, hash util class. Also added exceptions. The hibernation cfg file has been configured.

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,90 @@
+package mate.academy.model;
+
+import java.util.Arrays;
+import java.util.Objects;
+import javax.persistence.Entity;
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        User user = (User) o;
+        return Objects.equals(id, user.id)
+                && Objects.equals(email, user.email)
+                && Objects.equals(password, user.password)
+                && Arrays.equals(salt, user.salt);
+    }
+
+    @Override
+    public int hashCode() {
```
Rommelua, line 77: do you need these methods?

Rommelua (approved):

PR #480, opened 2023-03-25
first commit

dimonkoval:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("FROM User u WHERE u.email = :email", User.class)
+                    .setParameter("email", email).uniqueResultOptional();
+        }
```
aaananas, line 41: see common mistakes

aaananas (changes requested):

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        return userFromDB.get();
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userFromDB = userService.findByEmail(email);
+        User user = userFromDB.orElseThrow(() ->
+                new NoSuchElementException("Can't registered."));
+        userService.add(user);
+        return user;
```
olekskov, line 35: It's seems that use have reversed logic here.
We should register user only if such email wasn't used before.
So if we don't find user it's OK scenario.
If we already have such we should throw proper exception that we cannot perform registration with same email

olekskov (changes requested):

uliana-andrushchuk (approved):

PR #488, opened 2023-04-08
Hw hibernate user service hw

JRuiy:
I am not sure that is correct login method. How is it possible to do it with only one condition state?

### DIFF
--- a/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
@@ -1,0 +1,36 @@
+package mate.academy.service.impl;
+
+import java.util.Optional;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userByEmail = userService.findByEmail(email);
+        if (userByEmail.isPresent()) {
```
andriy-shulepa, line 21: fix common mistake
image
https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-user-service/add-user-service-checklist

andriy-shulepa (changes requested):

aaananas (approved):

PR #499, opened 2023-04-14
initial commit

kateryna-mykh:
User model was created, UserDao, UserService, AuthenticationService were implemented.

### DIFF
--- a/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
@@ -1,0 +1,39 @@
+package mate.academy.service.impl;
+
+import java.util.Optional;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Inject;
+import mate.academy.lib.Service;
+import mate.academy.model.User;
+import mate.academy.service.AuthenticationService;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (!userFromDbOptional.isEmpty() && user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can't authenticate user with email " + email);
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        User user = new User();
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("User with such email already exist");
+        }
+        user.setEmail(email);
+        user.setPassword(password);
+        return userService.add(user);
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (!userFromDbOptional.isEmpty() && user.getPassword().equals(hashedPassword)) {
+            return user;
```
Ivan95kos, line 24: Your version can throw NullPointerException
```suggestion
        if (userFromDbOptional.isPresent()
                && userFromDbOptional.get().getPassword()
                .equals(HashUtil.hashPassword(password, userFromDbOptional.get().getSalt()))) {
            return userFromDbOptional.get();
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
kateryna-mykh: Ok. I will change the condition, but in my opinion, such a solution in one line reads difficult.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
```
Ivan95kos, line 22: Проблема не в умові, а в тому що ти робиш перед умовою, ти витягуєш юзера без перевірки чи він там присутній.
Наприклад юзера немає в базі і в тебе буде наступна ситуація.
```suggestion
        User user = userFromDbOptional.get(); метод get() поверне null і буде user = null
        String hashedPassword = HashUtil.hashPassword(password, user.getSalt()); тут ти кажеш null.getSalt() буде помилка.
```

Ivan95kos (changes requested):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
kateryna-mykh: > Your version can throw NullPointerException

Do you mean that if all the same would be in a condition the NullPointerException wouldn't be?

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
kateryna-mykh: Do you mean that if all the same would be in a condition the NullPointerException wouldn't be?

olekskov (approved):
Good job. 
Ref your comment: correct, now if user is not found then .findByEmail() method returns empty optional and userFromDbOptional.isPresent() check has false result. So we're jumping to throw new AuthenticationException on line 25 and no NPE is met.

PR #502, opened 2023-04-18
implement user dao and auth service

yurihhhT:


mishok321 (approved):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userOptional = userService.findByEmail(email);
+        User user = userOptional.get();
+        if (user == null
```
Andriy-dz, line 22: do we need a null check here?

PR #503, opened 2023-04-18
Authentication and User services implemented!

AnonimToster:


KyslyiViacheslav (approved):
perfect

src/main/java/mate/academy/Main.java
```diff
@@ -1,55 +1,37 @@
...
-        firstCinemaHall.setCapacity(100);
-        firstCinemaHall.setDescription("first hall with capacity 100");
-
-        CinemaHall secondCinemaHall = new CinemaHall();
-        secondCinemaHall.setCapacity(200);
-        secondCinemaHall.setDescription("second hall with capacity 200");
-
-        CinemaHallService cinemaHallService = null;
-        cinemaHallService.add(firstCinemaHall);
-        cinemaHallService.add(secondCinemaHall);
```
nacenik, line 32: Don't delete all this logic too. it helps you test, that other parts of your program work correctly. You should only add code to test a new logic

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password)
+            throws AuthenticationException {
+        Optional<User> user = userService.findByEmail(email);
+        if (!user.isPresent()
```
nacenik, line 22: Instead of `!user.isPresent()` use `user.isEmpty()`

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+                                user.get().getSalt()))) {
+            throw new AuthenticationException("Wrong email or login!");
+        }
+        return user.get();
+    }
+
+    @Override
+    public User register(String email, String password)
+            throws RegistrationException {
+        if (email.isEmpty() || password.isEmpty()) {
```
nacenik, line 33: what happens if the email or password is `null`?

nacenik (changes requested):
Good solution! Let’s improve it 😉

KseniiaMakarova (approved):

PR #515, opened 2023-04-19
Budkevych

De1eF:


src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,20 @@
...
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
 
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.academy");
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        MovieService movieService = (MovieService) injector.getInstance(MovieService.class);
```
IvanZuravlev228, line 18: I think will be better if you will make this variable, variable of class

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+            throw new DataProcessingException("Can't insert user " + user, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
```
IvanZuravlev228, line 42: You can use just HQL query, but this is optional

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+            Predicate predicate = criteriaBuilder.equal(root.get("email"), email);
+            criteriaQuery.where(predicate);
+            return Optional.of(session.createQuery(criteriaQuery).getSingleResult());
+        } catch (Exception e) {
+            return Optional.empty();
+        }
+    }
+
+    @Override
+    public List<User> getAll() {
```
IvanZuravlev228, line 57: in the task condition it is not written that you need to create getAll. It says that UserService should have two methods add and fingByEmail, and since UserService uses userDao, this method should not be

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import mate.academy.service.UserService;
+import mate.academy.util.SaltUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
```
IvanZuravlev228, line 18: I think this method can be done with one exception throwing

src/main/java/mate/academy/util/SaltUtil.java
```diff
@@ -0,0 +1,27 @@
...
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
IvanZuravlev228, line 23: You should throw exception with a  information message

src/main/java/mate/academy/util/SaltUtil.java
```diff
@@ -0,0 +1,27 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class SaltUtil {
```
IvanZuravlev228, line 6: This is not SaltUtil, this is HashUtil, because you make hash and not salt here. Also, maybe you need a method which will random generate the salt

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        throw exception;
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("User with this email already exists");
+        }
+        byte[] salt = new byte[2];
```
IvanZuravlev228, line 33: Will be better if you will use your new method that randomly generates salt

IvanZuravlev228 (changes requested):

IvanZuravlev228 (approved):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        AuthenticationException exception =
+                new AuthenticationException("Login or password is incorrect");
```
karma-o, line 21: Why do we need a variable?

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,27 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class HashUtil {
```
karma-o, line 6: this class also needs a `getSalt()` method, to be used in user service

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
karma-o: yes, first of all, we need to move the password hashing to the user service method as it is said in common mistakes, and we should use a HashUtil method, and bytes should be random and not hardcoded

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        if (userOptional.isPresent()
+                && userOptional.get().getPassword()
+                        .equals(HashUtil.hashPassword(password, userOptional.get().getSalt()))) {
+            return userOptional.get();
+        }
+        throw exception;
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
```
karma-o, line 32: add parameters null and empty checks

karma-o (changes requested):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -17,31 +17,27 @@ public class AuthenticationServiceImpl implements AuthenticationService {
...
+        throw new AuthenticationException("Login or password is incorrect");
     }
 
     @Override
     public User register(String email, String password) throws RegistrationException {
+        if (email == null || password == null) {
+            throw new RegistrationException(
+                    "An internal error has occurred, email or password values are null");
+        }
+        if (email.equals("") || password.equals("")) {
```
karma-o, line 35: use `isEmpty()` or `isBlank()` instead

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -18,7 +19,13 @@ public Optional<User> findByEmail(String email) {
     }
 
     @Override
-    public User add(User user) {
```
karma-o: method signature should have still been the same as it was defined in the task description, this user instance would just be set a salt and a new password, user object itself should still be created in the `register()` method

karma-o (changes requested):

karma-o (approved):

PR #528, opened 2023-04-20
user-service

AleksandrLts:
add HashUtil for encrypting user's password
add UserDao, UserDaoImpl, UserService, UserServiceImpl classes and imlement them add AuthenticationService and AuthenticationServiceImpl for login and register user
add User in model package

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,68 @@
+package mate.academy.model;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
+    private String password;
+    private byte[] salt;
+
+    public User() {
+    }
+
+    public User(String email, String password) {
+        this.email = email;
+        this.password = password;
+    }
+
+    public Long getId() {
+        return id;
+    }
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
EliSoll, line 15: our email field have to be unique - use @Column(unique = true) with email or @UniqueConstraint in general table annotation

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
```
EliSoll, line 64: we can't show sensitive data here like user's password and salt's value

EliSoll (approved):
pay attention to your user's toString() method and email's field there

src/main/java/mate/academy/model/User.java
karma-o: fix that and you are all good

karma-o (approved):

src/main/java/mate/academy/model/User.java
AleksandrLts: > fix that and you are all good

Already fixed in @Table annotation

PR #535, opened 2023-04-22
configuration file is added

metelska-anastasiia:
created User, UserDaoImpl and UserServiceImpl classes with implementations 
in AuthenticationServiceImpl class was added logic to register and login user 
Two new custom exceptions were created
HashUtil class was added to protect Users' password

### DIFF
--- a/src/main/java/mate/academy/exception/AuthenticationException.java
+++ b/src/main/java/mate/academy/exception/AuthenticationException.java
@@ -1,0 +1,7 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends Exception {
+
+    public AuthenticationException(String message) {
+    }
+}
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,59 @@
+package mate.academy.model;
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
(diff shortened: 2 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public void setId(Long id) {
+        this.id = id;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String login) {
```
olekskov, line 33: The parameter name should be 'email' instead of 'login' to match the field name and avoid confusion.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] salt = new byte[16];
+        secureRandom.nextBytes(salt);
+        return salt;
+    }
+
+    public static String hushPassword(String password, byte[] salt) {
```
olekskov, line 20: Typo in the method name: 'hushPassword' should be 'hashPassword' 🧐

src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends Exception {
+
+    public AuthenticationException(String message) {
+    }
```
olekskov, line 6:
```suggestion
    public AuthenticationException(String message) {
        super(message);
    }
```

olekskov (approved):
There are some typos but fine overall. Approved

PR #546, opened 2023-05-09
Solved

arthursilver111:


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public void register(String email, String password) throws RegistrationException {
+        Optional<User> userFomDbByLogin = userService.findByEmail(email);
+        if (password.length() <= 5) {
+            throw new RegistrationException("Password should be more then 5 symbols");
+        }
+        if (userFomDbByLogin.isPresent()) {
+            throw new RegistrationException("Can not register user with email: " + email);
+        }
+        User user = new User(email, password);
```
olekskov, line 26: Before adding the user, set the salt and hash the password

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Inject
+    private UserDao userDao;
+
+    @Override
+    public User add(User user) {
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
```
olekskov, line 20: this logic is more appropriate for Authentication service

src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends Exception {
```
olekskov, line 3: Consider making the AuthenticationException a RuntimeException, so that it does not need to be explicitly caught or declared in method signatures.

olekskov (approved):
Good job. Sorry for some misleading comments. Technical task conditions are more important than personal opinion :)

PR #566, opened 2023-06-26
task solution implemented

Tupadzaky:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -0,0 +1,10 @@
+package mate.academy.service;
+
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserService {
+    User add(User user);
+
+    Optional<User> findByLogin(String email);
```
MykytaKat, line 9: rename to 'findByEmail' for a better understanding

MykytaKat (approved):
great job! left only one minor comment

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "email='" + email + '\''
```
maxline, line 59: you can also add info about `id` value

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        Optional<User> userFromDb = userService.findByEmail(email);
+        if (userFromDb.isEmpty()) {
+            throw new AuthenticationException("This email has not been registered yet");
+        }
+        User user = userFromDb.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Couldn't login. Email or password is incorrect");
```
maxline, line 29: Could you try refactor this code in functional style?
You can use   method for `Optional` like `. filter` and `.orElseThrow` and get rid of using `isEmpty()` and `get()`

maxline (changes requested):
Very good job! Left a couple of minor suggestions

PR #569, opened 2023-06-30
hibernate user serviece hw

kcasPilot:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+    private static final int SALT_ARRAY_SIZE = 16;
+
+    private HashUtil() {
+
+    }
+
+    public static byte[] getSalt() {
+        byte[] salt = new byte[SALT_ARRAY_SIZE];
+        new SecureRandom().nextBytes(salt);
```
olekskov, line 17: You are creating a new instance of SecureRandom for each use. This is unnecessary and can be inefficient as creating a new SecureRandom instance can be expensive. You should create a single instance and reuse it.

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByEmail(email).isEmpty()) {
+            User user = new User();
+            user.setEmail(email);
+            user.setPassword(password);
+            return userService.add(user);
+        }
+        throw new RegistrationException("Can't register user with email " + email);
```
olekskov, line 37: RegistrationException message is not informative. Add more details about possible reasons of the error.

olekskov (approved):
Good job, approved.
Minor comments left to be considered further

src/main/java/mate/academy/util/HashUtil.java
kcasPilot: Ok, thank you )

PR #572, opened 2023-07-04
Solution

MTM1963:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
kozhukhovsky, line 13: image

kozhukhovsky (changes requested):

src/main/java/mate/academy/model/User.java
MTM1963: How should I do that?

MTM1963: Done.

kozhukhovsky (approved):

PR #578, opened 2023-07-06
solution for user-service-hw

vgladkyi:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORIThM = "SHA-512";
+
+    public HashUtil() {
+    }
+
```
okuzan, line 12: image

okuzan (changes requested):

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,67 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", login='" + login + '\''
+                + ", password='" + password + '\'' + '}';
```
Rommelua, line 65: image

Rommelua (changes requested):
https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-user-service/add-user-service-checklist

src/main/java/mate/academy/model/User.java
okuzan: [Don’t add password and salt to the toString() method.] (https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-user-service/add-user-service-checklist)

okuzan (changes requested):
Remove comments before committing

src/main/java/mate/academy/model/User.java
vgladkyi: main method contains comments that shows that users were added to db via register and add methods.

Ivan95kos (approved):
Nice
Your code should be readable and without comments

PR #580, opened 2023-07-08
Added solution to the task:

nikolasdymchenko7:
- created User model
- created UserDao
- created UserService
- created AuthenticationService
- created my own checked exception: RegistrationException and  AuthenticationException
- created instance of AuthenticationService using injector and tested all methods from it

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -60,8 +60,6 @@ public class User {
         return "User{"
                 + "id=" + id
                 + ", email='" + email + '\''
-                + ", password='" + password + '\''
-                + ", salt=" + Arrays.toString(salt)
                 + '}';
     }
 }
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
aaananas, line 14: see common mistakes

aaananas (changes requested):

src/main/java/mate/academy/model/User.java
nikolasdymchenko7: Sorry, fixed.

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,64 @@
...
```
aaananas, line 13:
```suggestion
    @Column(unique = true)
    private String email;
```

aaananas (approved):

PR #584, opened 2023-07-10
Solved

Poloutinho:


src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,22 @@
...
 public class Main {
-    public static void main(String[] args) {
-        MovieService movieService = null;
+    private static final Injector injector = Injector.getInstance("mate.academy");
+
+    public static void main(String[] args) throws RegistrationException, AuthenticationException {
+        MovieService movieService = (MovieService) injector.getInstance(MovieService.class);
 
         Movie fastAndFurious = new Movie("Fast and Furious");
         fastAndFurious.setDescription("An action film about street racing, heists, and spies.");
```
Gamemod13, line 23: Please use constructor with parameter, instead of default constructor with setters where it is possible.
You can read a good explanation here https://stackoverflow.com/questions/19359548/setter-methods-or-constructors
Fix all similar cases in the code.

Gamemod13 (changes requested):
Good Job! Let's make it better. And use more informative commit message, please ;)

src/main/java/mate/academy/Main.java
rovein: @Poloutinho it is up to you what option to choose. Here it is correct both the setter and constructor-based approaches

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<User> query = session.createQuery(
+                            "from User u where u.email = :email ", User.class)
+                    .setParameter("email", email);
+            return query.uniqueResultOptional();
```
rovein, line 43:
```suggestion
            return session.createQuery("from User u where u.email = :email ", User.class)
                    .setParameter("email", email)
                    .uniqueResultOptional();
```
Could be inlined.

rovein (changes requested):
Great job, but I don't see your `AuthenticationServiceImpl` class.

sarakhmen (approved):
good job :)

PR #586, opened 2023-07-10
hibernate-user-service-hw

eugenetolkunov:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
vika-trach, line 15: I think we need to use this annotation with email @Column(unique = true), because the email must be unique

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        Optional<User> userOptional = userService.findByEmail(email);
+        if (userOptional.isEmpty()) {
+            throw new AuthenticationException("Can`t authenticate user");
+        }
+        User user = userOptional.get();
+        String hashPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can`t authenticate user");
```
vika-trach, line 29: I'd recommend also adding a more descriptive message

vika-trach (changes requested):
Good job! Let's slightly improve the solution

vika-trach (approved):

src/main/java/mate/academy/model/User.java
rovein: Agree with @vika-trach, let's do it.

rovein (changes requested):
Great job, but left one comment, check first point in the checklist.

rovein (approved):
Great job!

PR #589, opened 2023-07-10
Looks like everything workin good, checkstyle done

MykNich:


src/main/java/mate/academy/Main.java
```diff
@@ -2,17 +2,23 @@
...
 import mate.academy.service.MovieSessionService;
 
 public class Main {
-    public static void main(String[] args) {
-        MovieService movieService = null;
+    private static final Injector injector = Injector.getInstance("mate.academy");
 
+    public static void main(String[] args) throws RegistrationException, AuthenticationException {
+        MovieService movieService =
+                (MovieService) injector.getInstance(MovieService.class);
```
ViktorBykh, line 21: Hereinafter it's better to move instances:
```suggestion
    private static final Injector injector = Injector.getInstance("mate.academy");
    private static final MovieService movieService =
                (MovieService) injector.getInstance(MovieService.class);

    public static void main(String[] args) throws RegistrationException, AuthenticationException {
```

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<User> query = session.createQuery(
+                    "SELECT u FROM User u WHERE u.email = :email", User.class);
+            query.setParameter("email", email);
+            return Optional.ofNullable(query.uniqueResult());
```
ViktorBykh, line 43: Here we could use .uniqueResultOptional();
```suggestion
            return query.uniqueResultOptional();
```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,59 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
```
ViktorBykh, line 56: We shouldn't add 'email' and 'password' in toString() method:
```suggestion
```

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userOptional = userService.findByEmail(email);
+        if (userOptional.isEmpty()) {
+            throw new AuthenticationException("There is no such email: " + email);
+        }
```
ViktorBykh, line 23: In this case, we should add less specifics, because it is more secure:
```suggestion
        if (userOptional.isEmpty()) {
            throw new AuthenticationException("Wrong login or password");
        }
```

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        if (user.getPassword().equals(HashUtil.hashPassword(password, user.getSalt()))) {
+            return user;
+        }
+        throw new AuthenticationException("Password is incorrect!");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("This email " + email + " is already present");
```
ViktorBykh, line 34: Same story: this approach is less secure
```suggestion
        if (userService.findByEmail(email).isPresent()) {
            throw new RegistrationException("Wrong login or password");
```

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        User user = new User();
+        user.setSalt(HashUtil.getSalt());
+        user.setEmail(email);
+        user.setPassword(HashUtil.hashPassword(password, user.getSalt()));
+
```
ViktorBykh, line 40:
```suggestion
```

ViktorBykh (changes requested):
Everything looks good)
I have a few suggestions

ViktorBykh (approved):
All looks great)

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userOptional = userService.findByEmail(email);
+        if (userOptional.isEmpty()) {
```
maxline, line 21: Let's use .orElseThrow() method when working with optional instead of using isEmpty() + get()

maxline (changes requested):
Good job! Let's improve the solution a bit

rovein (approved):
Great job!

PR #592, opened 2023-07-10
solution hibernate-user-service

mykyta-udav:


ASnazar (approved):
Good job!

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> optionalUser = userService.findByEmail(email);
+        if (optionalUser.isEmpty()) {
+            throw new AuthenticationException("Email or password incorrect");
+        }
+        User user = optionalUser.get();
```
maxline, line 24: Let's use .orElseThrow() method when working with optional instead of using isEmpty() + get()

maxline (changes requested):
Good job! Let's improve the solution a bit

PR #607, opened 2023-07-12
done

vrnvv:


src/main/java/mate/academy/dao/UserDaoImpl.java
```diff
@@ -0,0 +1,47 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            Query<User> query = session.createQuery("from User where email = :email", User.class);
```
Ivan95kos, line 42: For the future, It is better to write all SQL commands in upper case
```suggestion
            Query<User> query = session.createQuery("FROM User WHERE email = :email", User.class);
```

Ivan95kos (approved):
Great job, with minor comments

PR #609, opened 2023-07-12
Added task realization

pkobevko:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+                new AuthenticationException("Can't authenticate user"));
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (hashedPassword.equals(user.getPassword())) {
+            return user;
+        }
+        throw new AuthenticationException("Can't authenticate user");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
```
MykytaKat, line 29: You can also add a check for an empty email and password

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
```
MykytaKat, line 16:
```suggestion

```

MykytaKat (approved):
Great solutison! Left only one minor comment

src/main/java/mate/academy/exception/RegistrationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class RegistrationException extends Exception {
```
maxline, line 3:
```suggestion
public class RegistrationException extends RuntimeException  {
```

src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends Exception {
```
maxline, line 3:
```suggestion
public class AuthenticationException extends RuntimeException  {
```

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +56,22 @@ public static void main(String[] args) {
...
-        MovieSessionService movieSessionService = null;
         movieSessionService.add(tomorrowMovieSession);
         movieSessionService.add(yesterdayMovieSession);
 
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
-                        fastAndFurious.getId(), LocalDate.now()));
+                fastAndFurious.getId(), LocalDate.now()));
+
+        try {
```
maxline, line 66: let's extends  AuthenticationException  and RegistrationException from RuntimeException, so you can get rid from try catch block here.

maxline (changes requested):
Good job! Let's improve the soluiton

src/main/java/mate/academy/exception/RegistrationException.java
maxline: Sorry, this suggestion was added by accident, in this task you should extends rom checked exception. The previous version of implementation was correct.

src/main/java/mate/academy/exception/AuthenticationException.java
maxline: @pkobevko Sorry, this suggestion was added by accident, in this task you should extends rom checked exception. The previous version of implementation was correct.

maxline (approved):
Approved. And I also noticed that you didn't fix the remarks from student review. Its is not critical, but please be more attentive in the future.

PR #617, opened 2023-07-31
created model User and DAO and Service for its processing

irynamatveieva:


src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,9 @@
+package mate.academy.exception;
+
+import java.io.IOException;
+
+public class AuthenticationException extends IOException {
```
Rommelua, line 5:
```suggestion
public class AuthenticationException extends Exception {
```

src/main/java/mate/academy/exception/RegistrationException.java
```diff
@@ -0,0 +1,9 @@
+package mate.academy.exception;
+
+import java.io.IOException;
+
+public class RegistrationException extends IOException {
```
Rommelua, line 5:
```suggestion
public class RegistrationException extends Exception {
```

Rommelua (approved):

PR #619, opened 2023-08-01
Hw branch

yaaariiik:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+                    + ". Email or password was incorrect.");
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can`t authenticate user with email " + email
+                + ". Email or password was incorrect.");
+    }
```
okuzan, line 32: let's refactor it to throw just 1 exception

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+            return hashedPassword.toString();
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Could not create hash using SHA-512 algorithm", e);
```
okuzan, line 31:
```suggestion
            throw new RuntimeException("Could not create hash using " + CRYPTO_ALGORITHM + "  algorithm", e);

```

okuzan (changes requested):

okuzan (approved):

PR #622, opened 2023-08-04
hibernate-user-service-hw36-v1.0-ready-to-check

Andrii-Zakharchenko-UA:


src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends RuntimeException {
```
Rommelua, line 3: image

Rommelua (changes requested):

PR #636, opened 2023-08-18
created a user model, created a dao service and authorization levels,…

0-zagar-0:
… created my own checked exceptions, and tested the methods in the main methodÐ


olekskov (approved):
Good job

PR #639, opened 2023-08-19
hw-hibernate-user-service

AntonZhdanov:


src/main/java/mate/academy/Main.java
```diff
@@ -44,10 +51,26 @@ public static void main(String[] args) {
...
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        User user = new User();
+        user.setEmail("TomCruzxxxx@gmail.com");
+        user.setPassword("TopGunMaverick@!");
+        try {
+            authenticationService.register(user.getEmail(), user.getPassword());
+            authenticationService.login(user.getEmail(), user.getPassword());
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Can't register user with email:" + user.getEmail()
+            + "and password: " + user.getPassword(), e);
```
JeniaSan, line 69: I think you mustn't show user's password here

JeniaSan, line 69: I think you mustn't show user's password here

JeniaSan (changes requested):
Very good solution, but only one minor mistake

nacenik (approved):
LGTM! 👍

PR #641, opened 2023-08-20
added solution

IhorKalaur:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,63 @@
...
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "users")
```
JeniaSan, line 11: Please use singular nouns for naming tables

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can't authenticate user with email " + email);
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (!user.getPassword().equals(hashedPassword)) {
+            throw new AuthenticationException("Can't authenticate user with email, " + email
+                    + ". The password is incorrect ");
+        }
+        return user;
```
JeniaSan, line 30: you can simplify this code. use two checks only in one "if" block

JeniaSan (changes requested):
good job, lets improve some minor mistakes

src/main/java/mate/academy/model/User.java
IhorKalaur: > Please use singular nouns for naming tables

Based on the general structure of the project, the names for other tables are used in the plural, I think it would be incorrect to use the singular for this case

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        User user = userService.findByEmail(email).orElseThrow(()
+                -> new AuthenticationException("Can't authenticate user with email " + email
```
andrii-hoienko, line 20: Better to simplify error message, just say that email or password is invalid. Also it is possible to use one check here

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
...
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    private HashUtil() {
+
+    }
+
+    public static byte[] getSalt() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] salt = new byte[16];
```
andrii-hoienko, line 16: magic number

andrii-hoienko (changes requested):
Good, but let`s improve your solution

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,63 @@
...
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+    @Column(unique = true)
+    private String email;
+    private String password;
+    private byte[] salt;
+
+    public User() {
```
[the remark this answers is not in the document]
IhorKalaur: > Is it required to create no args constructor manually?

One of the requirements of Entities is a constructor without parameters, I obviously wrote it just to play it safe.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -16,15 +16,16 @@ public class AuthenticationServiceImpl implements AuthenticationService {
...
     @Override
     public User login(String email, String password) throws AuthenticationException {
-        User user = userService.findByEmail(email).orElseThrow(()
-                -> new AuthenticationException("Can't authenticate user with email " + email
-                + ". There isn't in DB"));
-        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
-        if (!user.getPassword().equals(hashedPassword)) {
-            throw new AuthenticationException("Can't authenticate user with email, " + email
-                    + ". The password is incorrect ");
+        try {
```
vovasalyha, line 19: It's no a clear handling of the possible AuthenticationException, because you catch all exceptions which can give you false-positive result. Use your previous version but simply add in the `if` check `user.isEmpty()` and give it a general message.

vovasalyha (changes requested):
Good, but let's improve your login method logic.

vovasalyha (approved):
Nice👍

PR #642, opened 2023-08-20
implemented user 'register' and 'login' methods

DmytroV95:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,51 @@
...
+        return user;
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            return session.createQuery("FROM User "
+                            + "WHERE email = :email", User.class)
+                    .setParameter("email", email)
```
Denis-Balako, line 44: Recommend using a constant for parameter names

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "user")
```
Denis-Balako, line 11: Table can store multiple users, not only one

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface AuthenticationService {
+    User login(String email, String password) throws AuthenticationException;
+
+    /**
+    * We should register a new user. New user entity will contains the email and password
+    * @param email - user email. should be unique for each user
+    * @param password - user password
+    * @return new user instance
+    */
```
Denis-Balako, line 15: Delete unnecessary comments.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        User userByEmail = userService.findByEmail(email).orElse(null);
+        if (userByEmail != null && isUserPasswordValid(password, userByEmail)) {
```
Denis-Balako, line 20: That's bad practice to operate with null. Just use isEmpty() for optional instead.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        User newUser = new User();
+        newUser.setEmail(email);
+        newUser.setPassword(password);
+        return userService.add(newUser);
+    }
+
+    private static boolean isUserPasswordValid(String password, User userByEmail) {
+        String hashedPassword = HashUtil.hashPassword(password, userByEmail.getSalt());
+        return userByEmail.getPassword().equals(hashedPassword)
+                && !userByEmail.getPassword().isEmpty();
```
Denis-Balako, line 42: Function has unpredictable behavior. You need to rename the method, for example "isUserPasswordValidAndNotEmpty" or move this check to calling function

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,37 @@
...
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b: digest) {
+                hashedPassword.append(String.format(FORMAT_IDENTIFICATION, b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Could not create hash using "
+                    + CRYPTO_ALGORITHM + " algorithm",e);
```
Denis-Balako, line 33:
```suggestion
                    + CRYPTO_ALGORITHM + " algorithm", e);
```

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,37 @@
...
+        return salt;
+    }
+
+    public static String hashPassword(String password, byte[] salt) {
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b: digest) {
```
Denis-Balako, line 28:
```suggestion
            for (byte b : digest) {
```

Denis-Balako (changes requested):
Good solution! Let's improve it.

src/main/java/mate/academy/model/User.java
Denis-Balako: My bad, naming is OK.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
DmytroV95: If password not equals hashedPassword we return false, also if password is empty we also return false. In both these cases, the password is not valid. So I think method naming is appropriate.

DmytroV95:
image

nacenik (approved):
LGTM! 👍

PR #645, opened 2023-08-21
Implemented User model, UserDao, UserService and AuthenticationServic…

Someboty:
…e with implementations, RegistrationException and AuthenticationException

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,68 @@
+package mate.academy.model;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+    @Column(nullable = false, unique = true)
+    private String email;
+    @Column(nullable = false)
+    private String password;
+    @Column(nullable = false)
+    private byte[] salt;
+
+    public User() {
+    }
+
+    public User(String email, String password) {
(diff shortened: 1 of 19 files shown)


ConfusedSatlan (approved):

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
+package mate.academy.model;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
+public class User {
```
andrii-hoienko, line 10: It is good practive to add @Table annotation when defining entity

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new AuthenticationException("Can't authenticate user with email: " + email);
+        }
+        return userOptional.get();
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (email.isEmpty() || password.isEmpty()) {
+            throw new RegistrationException(
+                    "Can't register a user with empty email or password, email: " + email);
```
andrii-hoienko, line 31: Why do you show only email here, show password as well because error message corresponds to both email and password

andrii-hoienko (changes requested):
Good, but let`s improve your solution

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
Someboty: But it's sensitive information, should we save it?

Vasul-Pankiv (approved):

PR #652, opened 2023-08-27
completed the main task

ivan0dyatlyukkk:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +52,29 @@ public static void main(String[] args) {
...
                         fastAndFurious.getId(), LocalDate.now()));
+
+        //TESTING
+        AuthenticationService authenticationService = (AuthenticationService) INJECTOR
+                                                        .getInstance(AuthenticationService.class);
+        try {
+            User ivan = authenticationService.register("ivan@gmail.com", "qwerty1234");
+            User bob = authenticationService.register("bob@gmail.com", "password123456");
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Can't register the user: ", e);
```
ihorvasylenko23, line 71: instead of throw new RuntimeException("Can't register the user: ", e) it is better to use more specific exception like RegistrationException

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +52,29 @@ public static void main(String[] args) {
...
+        }
+        try {
+            System.out.println(authenticationService.login("ivan@gmail.com", "qwerty1234"));
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("Can't login the user: ", e);
```
ihorvasylenko23, line 76: instead of throw new RuntimeException("Can't login the user: ", e)  it is better to use more specific exception like AuthenticationException

ihorvasylenko23 (approved):
Looks good on me. Added one recommendation

PR #657, opened 2023-08-30
Develop

silchenko-arsen:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
aaananas, line 17: see common mistakes

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,90 @@
...
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt)
```
aaananas, line 87: password and ssalt should not be exposed

aaananas (changes requested):

kozhukhovsky (approved):

PR #660, opened 2023-09-01
Homework hibernate-user-service.

UniyatYura:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,60 @@
...
+            transaction = session.beginTransaction();
+            session.persist(user);
+            transaction.commit();
+            return user;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't save user " + user + " to DB , ", e);
+        }
```
olekskov, line 28:
```suggestion
        } finally {
            if (session != null) {
                session.close();
            }
        }
```

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            User user = new User();
+            user.setEmail(email);
+            user.setPassword(password);
+            userService.add(user);
+            return user;
+        }
+        throw new RegistrationException("Can't register user from email: " + email);
+    }
```
olekskov, line 42: Please consider switching of blocks within and outside if block.
This will not change how program works, it's already fine now.
But with my proposal code seems more obvious (we have validation, then main code)
Also more code lines have less tabulation level
```suggestion
    public User register(String email, String password) throws RegistrationException {
        Optional<User> userFromDbOptional = userService.findByEmail(email);
        if (userFromDbOptional.isPresent()) {
            throw new RegistrationException("Can't register user from email: " + email);
        }
        User user = new User();
        user.setEmail(email);
        user.setPassword(password);
        userService.add(user);
        return user;
    }
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #665, opened 2023-09-08
Solution of this task

IvanovYurii:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Injector INJECTOR = Injector.getInstance("mate.academy");
+    private static final UserService USER_SERVICE = (UserService) INJECTOR
+            .getInstance(UserService.class);
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
```
Ivan95kos, line 20: Знімок екрана 2023-09-08 о 18 07 26

src/main/java/mate/academy/Main.java
```diff
@@ -2,22 +2,39 @@
...
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
 
 public class Main {
-    public static void main(String[] args) {
-        MovieService movieService = null;
+    private static final Injector INJECTOR = Injector.getInstance("mate.academy");
+    private static final MovieService MOVIE_SERVICE = (MovieService) INJECTOR
+            .getInstance(MovieService.class);
```
Ivan95kos, line 22: In my opinion, it is bad practice to set the name for Services as a constant.
fix it everywhere
```suggestion
    private final Injector injector = Injector.getInstance("mate.academy");
    private final MovieService movieService = (MovieService) injector
            .getInstance(MovieService.class);
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/academy/Main.java
IvanovYurii: > In my opinion, it is bad practice to set the name for Services as a constant. fix it everywhere

image

Ivan95kos (approved):
Nice

PR #667, opened 2023-09-13
Hibernate user service hw

karvatskyi:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        return user;
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userFromDbByEmail = userService.findByEmail(email);
+        if (!userFromDbByEmail.isEmpty()) {
+            throw new RegistrationException("This user is registered");
+        }
```
olekskov, line 40: expressions with ! are harder to read. Better to avoid them when possible
```suggestion
        if (userFromDbByEmail.isPresent()) {
            throw new RegistrationException("This user is registered");
        }

```

olekskov (approved):
Good job

PR #671, opened 2023-09-15
Hibernate user service hw

vitaliiplys:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -0,0 +1,10 @@
+package mate.academy.service;
+
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserService {
+    User add(User user);
+
+    Optional<User> findByEmail(String email); // we will use this `Optional` later
```
okuzan, line 9: remove comments

okuzan (approved):

PR #677, opened 2023-09-21
hibernate-user-service solution added

Sergey-Bogatkin:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
kozhukhovsky, line 15: add annotation, what is field is unique

kozhukhovsky (approved):

PR #679, opened 2023-09-25
Hw hibernate user service hw

urlo0k:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
```
Ivan95kos, line 19: Знімок екрана 2023-09-26 о 17 42 00

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    public UserDaoImpl(SessionFactory sessionFactory) {
+        this.sessionFactory = sessionFactory;
+    }
+
+    public UserDaoImpl() {
+        this.sessionFactory = HibernateUtil.getSessionFactory();
+    }
+
+    @Override
+    public User add(User user) {
```
Ivan95kos, line 30: You should only use the try-with-resources with the read operation
Знімок екрана 2023-09-26 о 17 43 34

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    public HashUtil() {
```
okuzan, line 10:
```suggestion
    private HashUtil() {
```

okuzan (changes requested):
Utility classes should not be instantiated

kozhukhovsky (approved):

PR #688, opened 2023-10-02
implemented user_service app

vsych321:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+
+@Dao
+public class UserDaoImpl implements UserDao {
+    @Override
+    public User add(User user) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
```
ystankevych, line 16: Do you really need to store SessionFactor local variable. getSessionFactory is static method and you need SessionFactory only to open your session (but it's up to you) (Such comment for other places =) )

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            session.save(user);
```
ystankevych, line 22: save is deprecated method. I think it would be better to use persist instead
```suggestion
            session.persist(user);
```

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new RuntimeException("Can't add user to DB: " + user, e);
```
ystankevych, line 28:
```suggestion
            throw new DataProcessingException("Can't add user to DB: " + user, e);
```
just create another constructor for this exception and use it =)

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return user;
```
ystankevych, line 34: you could return user immediately from try block
```suggestion
       
```

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public User get(Long id) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            return session.get(User.class, id);
+        }
```
ystankevych, line 42: Why without catch block? This is from common mistakes - "Remember to add catch blocks for operations of all types on DAO layer."
```suggestion
        } catch(Exception e) {
        throw new DataProcessingException(your message, cause);
        }
```

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public Optional<User> findByLogin(String login) {
+        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        try (Session session = sessionFactory.openSession()) {
+            return session.createQuery(
+                    "from User u where u.login = :login", User.class)
+                    .setParameter("login", login)
+                    .uniqueResultOptional();
+        }
```
ystankevych, line 53: Why without catch block? This is from common mistakes - "Remember to add catch blocks for operations of all types on DAO layer."
```suggestion
        } catch(Exception e) {
        throw new DataProcessingException(your message, cause);
        }
```

ystankevych (changes requested):
Hi! Great job! I just left some comments

PR #697, opened 2023-10-16
add homework

dmytro-voronin:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can`t authenticate user");
+        }
```
okuzan, line 23: use one if, throw exception once

okuzan (changes requested):

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,75 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email
+                + '\'' + ", password='" + password
```
Rommelua, line 72: image

Rommelua (changes requested):

olekskov (approved):

PR #704, opened 2023-10-27
user service was implemented

AndreyPolezhaiev1:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,59 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
```
sokimaaa, line 56: I think, It is bad idea to show sensitive data.
```suggestion
                + ", password='" + password + '\''

```

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        User user = userFromDbOptional.get();
```
sokimaaa, line 21: never use Optional#get without previuos null-check, e.g. Optional#isPresent

Consider using Optional#or, Optional#orElseThrow, ...

sokimaaa (changes requested):

src/main/java/mate/academy/model/User.java
AndreyPolezhaiev1: Я думал, что убрал

kerrrusha (approved):

PR #706, opened 2023-10-31
Implemented the requirement of the task

flekros:


src/main/java/mate/academy/AuthenticaonException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy;
+
+public class AuthenticaonException extends Exception {
```
mnyshenko, line 3: minor: Authentication*

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        throw new AuthenticationException("Password is incorrect");
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (password.isEmpty() || email.isEmpty()) {
+            throw new RegistrationException("Incorrect input params!"
+                    + " Email or password can't be empty!");
+        }
+        if (userService.findByEmail(email).isEmpty()) {
```
mnyshenko, line 34: `userService.findByEmail(email).isEmpty()` if this statement is true we should not throw any exceptions because it means that there is no user with such an email so we can register a new one

mnyshenko (changes requested):
Good job, but there is 1 bug to be fixed

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        User user = userService.findByEmail(email).orElseThrow(()
+                -> new AuthenticationException("User with email: " + email + " doesn't exist"));
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Password is incorrect");
```
mnyshenko, line 25: image
https://github.com/mate-academy/hibernate-user-service-hw/blob/master/checklist.md#common-mistakes-hibernate-user-service-hw

mnyshenko (approved):
Approved, but left one more comment to consider

PR #710, opened 2023-11-10
done

artificialLeo:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
+package mate.academy.model;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import java.util.Objects;
+
+@Entity
+@Table(name = "users")
```
kerrrusha, line 9: DB tables should be in singular, not plural
```suggestion
@Table(name = "user")
```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,68 @@
...
+public class User {
+
+    @Id
+    private String email;
```
kerrrusha, line 13: never forget about regular `id` field
```suggestion
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;    
    
    @Column(unique = true)
    private String email;
```

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        User user = userDao.findByEmail(email)
+                .orElseThrow(() -> new AuthenticationException("Authentication failed."
+                        + " User not found."));
+
+        byte[] storedSalt = HashUtil.generateSalt();
+        byte[] hashedPassword = HashUtil.hashPassword(password, storedSalt);
+
+        if (!Arrays.equals(hashedPassword, user.getPassword())) {
```
kerrrusha, line 28: better extract it to the separate method

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,28 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+
+    private static final String HASH_ALGORITHM = "SHA-512";
+
```
kerrrusha, line 10: create private constructor for this Util class so it can't be instantiated mistakenly

kerrrusha (changes requested):
overall good job

kerrrusha (approved):
good job

PR #712, opened 2023-11-14
created model of User and UserDao and UserService implementations

dima-semeniuk:


src/main/java/mate/academy/security/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+            if (password != null) {
+                String hashedPassword = HashUtil.hashPassword(password, userFromDb.getSalt());
+                if (userFromDb.getPassword().equals(hashedPassword)) {
+                    return userFromDb;
+
+                }
+            }
+        }
+        throw new AuthenticationException("Email or password is incorrect");
+    }
```
HovorukhaBohdan, line 32: I think it's too complicated
```suggestion
        Optional<User> userFromDbOptional = userService.findByEmail(email);
        User userFromDb = userFromDbOptional.orElseThrow(() ->  new AuthenticationException("Email or password is incorrect"));
        if (password != null) { // I think password can't be null in DB, what do u think?
            String hashedPassword = HashUtil.hashPassword(password, userFromDb.getSalt());
            if (userFromDb.getPassword().equals(hashedPassword)) {
                return userFromDb;
            }
```

HovorukhaBohdan (approved):
Great job. Only one minor point to consider.

src/main/java/mate/academy/security/service/impl/AuthenticationServiceImpl.java
dima-semeniuk: If optional is not empty, then when I check if the hashes are equals, I will again need to throw exception if they are not equal, because otherwise the code will not compile (because all "if" constructions may not be executed). In my code I check all the conditions  step by step and throw an exception if something doesn't true. And if the password is null, then the method "hashPassword" will throw NPE, so I need to check it

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<User> query = session.createQuery("from User u where u.email = : email",
```
vovasalyha, line 40:
```suggestion
            Query<User> query = session.createQuery("from User u where u.email = :email",
```

vovasalyha (approved):
LGTM

PR #714, opened 2023-11-16
finished hw

DZarudnyi:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
```
kozhukhovsky, line 19: image

kozhukhovsky (changes requested):

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
DZarudnyi: So you suggesting me to do all those operations and only AFTER them - check if user is not in DB? Why should i use resources for that, when existing approach uses less resources?

DZarudnyi:
To check if passwords match, i need to get salt. Salt is added in 'add' method, so only way for me to get salt - is to get it from DB. So the first thing i need to do is to get Optional<User> from DB (which may be empty).
If I to combine those checks in one, i will get user optional from DB, try to get salt from that user and only then perform check. If this optional is null - I'll get an error. So it makes much more sense to perform empty check first and only then do all the other operations.

If I'm wrong - please explain to me how.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
okuzan: Hashing isn't a particularly resource-intensive operation. Anyway, we don't need to getSalt of empty user optional
Also, while having a different message is more user-friendly, it's less secure, since we give clues to a potential attacker
```
    if (userCheck.isEmpty() || !passwordValidation(userCheck.get(), password)) {
        throw new AuthenticationException("Wrong login or password!");
    }
```

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,37 @@
...
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPas.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(
+                    "Was not able to create hashed password with SHA-512 algorithm", e);
```
okuzan, line 32: use your constant in the message

okuzan (changes requested):

okuzan (approved):

PR #719, opened 2023-11-19
Add task implementation.

MinoDentori:
1. Create a HashUtil class to store methods for password hashing.
2. Created custom Registration and Authentication exception classes for specific cases in AutehntificationServiceImp.
3. Created AutehntificationService interface and AutehntificationServiceImp class (registration, login).
4. Created User entity, UserDao, and UserService, with specific (findByEmail method). 
5. Add user creation, finding tests in Main. 

For now, the goal of the last test is to get RegistrationException (Can't register user. Email is already used), so pay attention to this case.

### DIFF
--- a/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
@@ -1,0 +1,53 @@
+package mate.academy.service.impl;
+
+import java.util.Optional;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Inject;
+import mate.academy.lib.Service;
+import mate.academy.model.User;
+import mate.academy.service.AuthenticationService;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final String CANT_LOGIN_EXCEPTION_MESSAGE =
+            "Can't login. Login or password is wrong.";
+    private static final String EXISTED_EMAIL_EXCEPTION_MESSAGE =
+            "Can't register user. This email is already used";
+    private static final String EMPTY_EMAIL_EXCEPTION_MESSAGE =
+            "Email can't be empty";
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException(CANT_LOGIN_EXCEPTION_MESSAGE);
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException(CANT_LOGIN_EXCEPTION_MESSAGE);
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        if (password.isEmpty()) {
+            throw new RegistrationException(EMPTY_EMAIL_EXCEPTION_MESSAGE);
+        }
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException(EXISTED_EMAIL_EXCEPTION_MESSAGE);
+        }
+        User user = new User();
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException(CANT_LOGIN_EXCEPTION_MESSAGE);
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException(CANT_LOGIN_EXCEPTION_MESSAGE);
```
kshuryhin, line 35: Create only one condition for throwing `AuthenticationException` in `login()` method. You may combine two checks: whether the user has been found by login and does passwords match.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        if (password.isEmpty()) {
+            throw new RegistrationException(EMPTY_EMAIL_EXCEPTION_MESSAGE);
+        }
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException(EXISTED_EMAIL_EXCEPTION_MESSAGE);
+        }
+        User user = new User();
+        user.setSalt(HashUtil.getSalt());
+        user.setEmail(email);
+        user.setPassword(HashUtil.hashPassword(password, user.getSalt()));
```
kshuryhin, line 49: You should create salt and hash password in `UserService` `add()` method.

kshuryhin (changes requested):
Good job, but there are some points to consider

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
MinoDentori: Tried to fix it, but I made the implementation in this way because we don't have a password in the signature of the method in UserService, so I guess that we need to create a hash, where we have a password. For now, I set a password with no hash for the registration time, and changed it to the hash version in UserService.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
MinoDentori: I tried to fix it. If I understand right we can use || in if in a way, that if we have no user, we will throw an exception, or if a user exists, we can get the user from userOptional and be sure that we won't get NPE. For now, it looks overwhelmed by text, any way, to make it better?

Andriy-Hrynevetskyi:
I don't know why i can review this after a mentor )))) byt everything looks fine)/

Andriy-Hrynevetskyi (approved):
Gj mate). A little advice next time hit the "ready to check" button on a platform but not here on Git)

MinoDentori:
> I don't know why i can review this after a mentor )))) byt everything looks fine)

I guess this situation possible, because this pr didn't have approve from student. It was sent to mentors after 24 hours automatically.

MinoDentori:
> Gj mate). A little advice next time hit the "ready to check" button on a platform but not here on Git) 

Yeah, I guess I was in autopilot mode and asked for approve from mentor on GitHub this time  😵‍💫

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()
+                || isPasswordCorrect(userFromDbOptional.get().getPassword(),
+                        HashUtil.hashPassword(password, userFromDbOptional.get().getSalt()))) {
+            throw new AuthenticationException(CANT_LOGIN_EXCEPTION_MESSAGE);
+        }
```
bhdnchui, line 31: Optional. We can separate this big check for two: 
1) check if the user not registered - inform user that no such user in database, 
2) if passwords are different - inform user that credentials are wrong.

It will simplify that code and it allow to make login process more clear for the user.

bhdnchui (approved):
Good job!

PR #720, opened 2023-11-21
AuthenticationService, first commit

Stelmaschyk:


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        user.setEmail(email);
+        user.setPassword(password);
+        try {
+            userService.add(user);
+            return user;
+        } catch (RuntimeException e) {
+            throw new RegistrationException("Can't register user with email: "
+                    + email + ", and password " + password, e);
+        }
+    }
```
okuzan, line 40: use checks like these instead of try catch
image

okuzan (changes requested):

kozhukhovsky (approved):

PR #722, opened 2023-11-21
Hibernate user service hw solution

BolsunovaNataliia:


src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can't authenticate user");
+        }
```
proxychang, line 22: Think about removing redudant throwing and throw only one exception. I mean you can change the condition that if userFromDbOptional is present, than return user, if not - thow AE

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        User user = new User();
+        user.setEmail(email);
+        user.setSalt(HashUtil.getSalt());
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        user.setPassword(hashedPassword);
+        if (user == null || email == null || password == null) {
```
proxychang, line 38: You're setting salt here before this condition, so even if email and/or password will be null, user won't be null

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+            throw new RegistrationException("Can't registrate user");
+        }
+        userService.add(user);
+        return user;
```
proxychang, line 42: You can simplify it

proxychang (changes requested):
Good job! Consider a few comments and it will help make your code a little better

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
BolsunovaNataliia: How?

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isEmpty()) {
+            throw new AuthenticationException("Can't authenticate user");
+        }
+        User user = userFromDbOptional.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Can't authenticate user");
```
kshuryhin, line 28: Create only one condition for throwing `AuthenticationException` in `login()` method. You may combine two checks: whether the user has been found by login and does passwords match.

kshuryhin (changes requested):
Good job, just minor point to consider

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession(); ) {
+            return session.createQuery("from User u where u.email = :email", User.class)
```
bhdnchui, line 39: Use UPPERCASE letters here.

bhdnchui (changes requested):
Good job! Just a few points to consider.

bhdnchui (approved):
Good job!

PR #734, opened 2023-12-11
added authentication_service

Oleksii-Bilko:


src/main/java/mate/academy/exception/RegistrationException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.exception;
+
+public class RegistrationException extends Exception {
```
kerrrusha, line 3: prefer unchecked exceptions over checked
```suggestion5
public class RegistrationException extends RuntimeException {
```

kerrrusha (approved):

PR #743, opened 2024-01-09
Add solution

slaybrute:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
kozhukhovsky, line 19: unique?

kozhukhovsky (changes requested):

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public User add(User user) {
+        return userDao.add(user);
+    }
+
+    @Override
+    public User get(Long id) {
+        return userDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Cannot find user by id: " + id));
```
okuzan, line 23: DataProcessing is for DAO level, throw EntityNotFound on service level everywhere instead

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,36 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+
```
okuzan, line 8:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -0,0 +1,11 @@
...
+
+import mate.academy.model.User;
+
+public interface UserService {
+    User add(User user);
+
+    User get(Long id);
+
+    User get(String email);
+}
```
okuzan, line 11: image
here, service should return optional
it's expected situation - not to be able to find a user by email, we shouldn't throw exception

okuzan (changes requested):

kozhukhovsky (approved):

PR #745, opened 2024-01-09
First solution of hibernate-user-service-hw

QbaSekowski:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
Rommelua, line 15: image

Rommelua (approved):
Good job, just add **unique=true** to the email field

PR #748, opened 2024-01-14
need some help

Kalixone:
"register method works, but the other one does not, i think it's because of hashing i can't compare them properly, they change dynamically i don't know how to fix it, need help"

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,64 @@
+package mate.academy.model;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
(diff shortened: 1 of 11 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,64 @@
...
+    public String toString() {
+        return "User{"
+                + "id="
+                + id
+                + ", email='"
+                + email
+                + '\''
+                + ", password='"
+                + password
+                + '\''
```
Rommelua, line 61: image

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+        String hashedPassword = HashUtil.hashPassword(password, salt);
+
+        user.setSalt(salt);
+        user.setPassword(hashedPassword);
+
+        User registeredUser = userService.add(user);
+
+        if (registeredUser == null) {
+            throw new RegistrationException("Failed to register user");
+        }
```
Rommelua, line 54:
```suggestion
```

Rommelua (changes requested):

Rommelua (approved):
good job

PR #757, opened 2024-01-26
hibernate user service solution

furthernull:


src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new AuthenticationException(
+                    "Can't authenticate user by email: " + email);
+        }
+        User user = userByEmail.get();
+        String hashedPassword = HashUtil.hashPassword(password, user.getSalt());
+        if (user.getPassword().equals(hashedPassword)) {
+            return user;
+        }
+        throw new AuthenticationException("Incorrect password");
+    }
```
olekskov, line 31: 1) It's not good to differentiate error message of wrong email from wrong password in order not to give understanding to potential fraudster if he is correct with some data already
2) Point 1 lead to potential code simplification
```suggestion
        Optional<User> user = userService.findByEmail(email);
        if (user.isPresent() && user.get().getPassword().equals(hashedPassword)) {
            return user.get();
        }
        throw new AuthenticationException("Incorrect login or password");
    }
```

olekskov (approved):
Good job overall, approved.
Please pay attention to comment below for further

PR #791, opened 2024-03-22
hw

AGuliaiev:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Could not create hash using SHA-512 algorithm", e);
```
okuzan, line 30: use your constant

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    public User save(User user) {
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
+        return userDao.save(user);
+    }
+
+    @Override
+    public Optional<User> findByLogin(String login) {
+        return userDao.findByLogin(login);
+    }
```
okuzan, line 33: service should return concrete object, not optional of object, use orElseThrow + entity NotFound + message +id, fix everywhere

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,67 @@
...
+    public void setSalt(byte[] salt) {
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
```
okuzan, line 64: do you think this is a good idea? we could leak security info
it's a checklist of common mistakes issue
https://github.com/mate-academy/hibernate-user-service-hw/blob/master/checklist.md

okuzan (changes requested):
security issue + unwrap optional

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
+        return userDao.save(user);
+    }
+
+    @Override
+    public Optional<User> findByLogin(String login) {
+        return Optional.ofNullable(userDao.findByLogin(login)
+                .orElseThrow(() ->
+                        new EntityNotFoundException("User with login " + login + " not found")));
+    }
```
olekskov, line 36:
```suggestion
    @Override
    public User findByLogin(String login) {
        return userDao.findByLogin(login)
                .orElseThrow(() ->
                        new EntityNotFoundException("User with login " + login + " not found"));
    }
```

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new DataProcessingException("Could not create hash using SHA-512 algorithm", e);
```
olekskov, line 31:
```suggestion
            throw new DataProcessingException("Could not create hash using " + CRYPTO_ALGORITH + " algorithm", e);
```

olekskov (changes requested):
Please fix issued highlighted in previous review.
Examples are below

olekskov (approved):

PR #797, opened 2024-04-01
add and implement userDao and userService

Bohadan-Markatov:


src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public User login(String login, String password) throws AuthenticationException {
+        User user = userService.findByLogin(login).orElseThrow(()
+                -> new AuthenticationException("This login: " + login + " doesn't exist!"));
+        if (user.getPassword().equals(HashUtil.hashPassword(password, user.getSalt()))) {
+            return user;
+        }
+        throw new AuthenticationException("Wrong password!");
+    }
```
SkillfulDev, line 25: To avoid brute force we don’t have to give a specific hint that of the data entered is incorrect. Also we have a rule 
image
```suggestion
    @Override
    public User login(String login, String password) throws AuthenticationException {
        User user = userService.findByLogin(login).orElseThrow(()
                -> new AuthenticationException("This login: " + login + " doesn't exist!"));
        if (user.getPassword().equals(HashUtil.hashPassword(password, user.getSalt()))) {
            return user;
        }
        throw new AuthenticationException("Wrong password!");
    }

```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,61 @@
...
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt)
```
SkillfulDev, line 58:
```suggestion
                
               
```
image

SkillfulDev (changes requested):
GJ !

SkillfulDev (approved):

PR #801, opened 2024-04-04
all features works

MolchanovAlexander:


src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,39 @@
...
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.User;
+import mate.academy.security.AuthenticationService;
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
 
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.academy");
```
TonyH277, line 18: not shure if this field should be final

src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,39 @@
...
     public static void main(String[] args) {
-        MovieService movieService = null;
+        AuthenticationService authenticationService = (AuthenticationService) injector
+                .getInstance(AuthenticationService.class);
+
+        try {
+            authenticationService.register("guf@dgigurda.com", "1234");
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Authentication failed", e);
+        }
```
TonyH277, line 28: exception will be thrown if it occurs from AuthService, and in those implementation catch block do the same and program will fail with Registration exception.
IMHO: we dont need try/catch here

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByLogin(String login) {
+        System.out.println("dao login : " + login);
+        Session session = getSession();
+        Transaction transaction = session.beginTransaction();
```
TonyH277, line 40: you don't need transaction in reading operations

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByLogin(String login) {
+        System.out.println("dao login : " + login);
```
TonyH277, line 38:
```suggestion
```

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        Session session = getSession();
```
TonyH277, line 39: you could use try with resourses here

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,71 @@
...
+        this.salt = salt;
+    }
+
+    @Override
+    public String toString() {
+        return "User {"
+                + "id='" + id + '\''
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
+                + ", salt='" + Arrays.toString(salt)
```
TonyH277, line 68:
```suggestion
```

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,16 @@
...
+        <property name="connection.password">---SQLdno12</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.User"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+    </session-factory>
+</hibernate-configuration>
```
TonyH277, line 16:
```suggestion
</hibernate-configuration>

```

TonyH277 (approved):
Nice work! Left some comments which actually my thoughts.

src/main/java/mate/academy/model/User.java
karma-o: agree, this is confidential info and should not be visible in any logs or other things like that
```suggestion
```

karma-o (changes requested):

karma-o (changes requested):
comment to reject on the platform

src/main/java/mate/academy/Main.java
MolchanovAlexander: Actually it works because injector is single and it's method calls from variable injector.

PR #804, opened 2024-04-07
User Service Implementation for Cinema Project

VolandevlodD:
This PR implements user service functionality for the Cinema project using Hibernate. Key points:

- Created User model and UserDAO.
- Implemented UserService and AuthenticationService interfaces.
- UserService methods: add(User), findByEmail(String).
- AuthenticationService methods: login(String, String) with AuthenticationException, register(String, String) with RegistrationException.
- Custom exceptions for error handling.
- Passwords hashed with salt.
- Tested in mate/academy/Main.main().

### DIFF
--- a/src/main/java/mate/academy/dao/impl/UserDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/UserDaoImpl.java
@@ -1,0 +1,51 @@
+package mate.academy.dao.impl;
+
+import java.util.Optional;
+import mate.academy.dao.UserDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.User;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class UserDaoImpl implements UserDao {
+    private final SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+
+    @Override
+    public User add(User user) {
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(user);
+            transaction.commit();
+            return user;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Could not add user: " + user, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = sessionFactory.openSession()) {
+            String hql = "from User where login = :email";
+            Query<User> findByEmailQuery = session.createQuery(hql, User.class).setParameter(
+                    "email", email);
+            return findByEmailQuery.uniqueResultOptional();
+        } catch (Exception e) {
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,51 @@
...
+            session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(user);
+            transaction.commit();
+            return user;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Could not add user: " + user, e);
```
nklimovych, line 32: I think it might be better to return the user ID instead of the entire user object

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    @Inject
+    private UserService userService;
+
+    @Override
+    public User login(String email, String password) throws AuthenticationException {
+        Optional<User> userFromDb = userService.findByEmail(email);
+        if (userFromDb.isPresent() && isValidPasswordForUser(userFromDb.get(), password)) {
```
nklimovych, line 22: Just a suggestion
```suggestion
        if (userFromDb.isPresent() && isPasswordValid(userFromDb.get(), password)) {
```

src/main/java/mate/academy/security/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userFromDb = userService.findByEmail(email);
+        if (userFromDb.isPresent()) {
+            throw new RegistrationException(String.format("User with email: %s  already exists",
+                    email));
+        }
+        if (password == null || password.isBlank() || password.length() < 6) {
+            throw new RegistrationException("Password must be at least 6 characters");
+        }
```
nklimovych, line 37: Maybe it would be better to add this validation to a separate method?

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,37 @@
...
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(HASH_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format(HEX_FORMAT, b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException(String.format(
+                    "Couldn't create hash using %s " + "algorithm!", HASH_ALGORITHM));
```
nklimovych, line 26:
```suggestion
                    "Couldn't create hash using %s algorithm!", HASH_ALGORITHM));
```

nklimovych (changes requested):
Great job! It looks great to me! Left you some suggestions 👍🏻

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
VolandevlodD: The `toString` method in the `User` class overrides only the `id` and `login` fields. Therefore, only the `id` and `login` will be printed. That's why I believe it's not a bad idea to return the entire `User` object because it provides a more informative message.

PR #809, opened 2024-04-11
implemented hibernate-user-service-hw

HAIOVYI:


TonyH277 (changes requested):
Nice! Left some comments to improve your solution

PR #811, opened 2024-04-13
Realized components for User (model, DAO, Service, AuthenticationExce…

Vlad1507:
…ption and RegistrationException). Configured hash utility class and hibernate configuration. Created an example in the Main class.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,0 +1,13 @@
+package mate.academy.service;
+
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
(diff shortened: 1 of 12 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -0,0 +1,13 @@
...
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.model.User;
+
+public interface AuthenticationService {
+
+    User login(String email, String password) throws AuthenticationException;
+
+    User register(String email, String password) throws RegistrationException;
+
```
MentorMA, line 12: image
Keep the same style

MentorMA (approved):

PR #814, opened 2024-04-15
Solution

ghost:


src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,31 @@
+package mate.academy.util;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
```
karma-o, line 7: check common mistakes, we should add **private** default constructor to the `HashUtil`, since it's a util class with static-only methods

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +58,18 @@ public static void main(String[] args) {
...
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
                         fastAndFurious.getId(), LocalDate.now()));
+
+        User newUser = new User();
+        newUser.setEmail("some@mail.com");
+        newUser.setPassword("somePassword");
+
+        userService.add(newUser);
+        System.out.println(userService.findByEmail("some@mail.com"));
```
karma-o, line 73: we should test `login()` and `register()` methods here

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -24,9 +24,7 @@ public class MovieSessionDaoImpl implements MovieSessionDao {
     @Override
     public MovieSession add(MovieSession movieSession) {
         Transaction transaction = null;
-        Session session = null;
-        try {
-            session = HibernateUtil.getSessionFactory().openSession();
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
```
karma-o, line 27: why did you do this?

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
```diff
@@ -36,10 +34,6 @@ public MovieSession add(MovieSession movieSession) {
                 transaction.rollback();
             }
             throw new DataProcessingException("Can't insert movie session" + movieSession, e);
-        } finally {
-            if (session != null) {
-                session.close();
-            }
```
karma-o, line 42: don't you need to close the session resources if we get some exception?
rollback these changes

karma-o (changes requested):

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
VlasenkoYevhenii: I didn't change the classes that I wasn't supposed to by the task.
Just forked and implemented stuff according to the README.md.

src/main/java/mate/academy/dao/impl/MovieSessionDaoImpl.java
VlasenkoYevhenii: Same here. I've seen some moments I didn't like but decided not to fix the things I wasn't asked to.

yevheniisynytsia (approved):

PR #821, opened 2024-05-06
solution

Shr1mpa:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,52 @@
...
+import jakarta.persistence.Table;
+import java.util.Arrays;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
Elena-Bruyako, line 16:
```suggestion
    @Column(nullable = false, unique = true)
    private String email;
```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,52 @@
...
+    private String password;
```
Elena-Bruyako, line 17:
```suggestion
    @Column(nullable = false)
    private String password;
```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,52 @@
...
+    private byte[] salt;
```
Elena-Bruyako, line 18:
```suggestion
    @Column(nullable = false)
    private byte[] salt;
```

src/main/java/mate/academy/dao/UserDao.java
```diff
@@ -0,0 +1,15 @@
+package mate.academy.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserDao {
+    User add(User user);
+
+    Optional<User> get(Long id);
```
Elena-Bruyako, line 10: You don't use these methods `Optional<User> get(Long id);` and `List<User> getAll();` so remove them

Elena-Bruyako (changes requested):

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,35 @@
...
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b: digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Could not create hash "
+                    + "using SHA-512 algorithm", e);
```
okuzan, line 31: use your constant here

okuzan (approved):

PR #835, opened 2024-06-08
remade_hw

Yam0r:


src/main/java/mate/academy/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,16 @@
+<hibernate-configuration>
```
okuzan, line 1: wrong location, should be src/main/resources

okuzan (changes requested):

PR #840, opened 2024-06-22
solution

GlebPashko:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        return user;
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            String hql = "from User where email = :email";
+            return session.createQuery(hql, User.class)
+                    .setParameter("email", email)
+                    .stream()
```
Elena-Bruyako, line 42:
```suggestion
                    .uniqueResultOptional());
```

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,63 @@
...
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private int id;
+    private String email;
```
Elena-Bruyako, line 13: Let's make the email unique. We need to check the email during user registration and make the email a unique field in the database.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        validateData(email, password);
+        if (userService.findByEmail(email).isPresent()) {
+            throw new RegistrationException("User with email " + email + " already exists");
+        }
+        User user = new User(email, password);
+        return userService.add(user);
+    }
+
+    public void validateData(String email, String password) throws AuthenticationException {
+        if (email.isEmpty() || password.isEmpty()) {
```
Elena-Bruyako, line 44: It's better to check is email and password not null also. 
isEmpty just checking that the length is 0

Elena-Bruyako (changes requested):

okuzan (approved):

PR #842, opened 2024-06-23
hibernate-user-service-hw-sokur-v1

MaksimSokur:
1. Create models: User
2. Create DAO User DAO
3. Create service UserService AuthenticationService
4. Create checked RegistrationException and AuthenticationException
5. Use salt in password hiding
6. In method main create instance and inject all methods

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,0 +1,47 @@
+package mate.academy.model;
+
+public class User {
+
+    private String email;
+    private String password;
+    private byte[] salt;
+
+    public User() {
+    }
+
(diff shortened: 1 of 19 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public User() {
+    }
+
+    public User(String email, String password) {
+        this.email = email;
+        this.password = password;
+    }
+
+    public User(String email, String password, byte[] salt) {
```
Elena-Bruyako, line 17: Do you use all these constructors?

Elena-Bruyako (changes requested):

PR #852, opened 2024-07-26
hw done

BelitskyiAlexandr:
could you explain why hashing is better done in a service? 
Isn't it logical that it is a one-time operation that is used only together with registration? 
Why we must separate logic that always works together?

### DIFF
--- a/src/main/java/mate/academy/security/AutheticationService.java
+++ b/src/main/java/mate/academy/security/AutheticationService.java
@@ -1,0 +1,17 @@
+package mate.academy.security;
+
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.model.User;
+
+public interface AutheticationService {
+    User login(String email, String password) throws AuthenticationException;
+
+    /**
+     * We should register a new user. The new user entity will contain the email and password
+     * @param email - user email. should be unique for each user
+     * @param password - user password
+     * @return new user instance
+     */
+    User register(String email, String password) throws RegistrationException;
+}
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -1,0 +1,10 @@
+package mate.academy.service;
+
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserService {
+    User add(User user);
+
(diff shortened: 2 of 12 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -0,0 +1,10 @@
+package mate.academy.service;
+
+import java.util.Optional;
+import mate.academy.model.User;
+
+public interface UserService {
+    User add(User user);
+
+    Optional<User> findByEmail(String email); // we will use this `Optional` later
```
sarakhmen, line 9:
```suggestion
    Optional<User> findByEmail(String email); // we will use this `Optional` later
```
please remove redundant comment

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,17 @@
...
+        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>
+        <property name="connection.username">root</property>
+        <property name="connection.password">rootroot</property>
+        <property name="show_sql">true</property>
+        <property name="format_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.model.CinemaHall" />
+        <mapping class="mate.academy.model.Movie" />
+        <mapping class="mate.academy.model.MovieSession" />
```
sarakhmen, line 14: why did you have these lines of code in this PR? They should have already been merged

src/main/java/mate/academy/security/AutheticationService.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface AutheticationService {
+    User login(String email, String password) throws AuthenticationException;
+
+    /**
+     * We should register a new user. The new user entity will contain the email and password
+     * @param email - user email. should be unique for each user
+     * @param password - user password
+     * @return new user instance
+     */
```
sarakhmen, line 15: let's remove this single javadoc

sarakhmen (changes requested):
let's improve your solution :) As for your question it seems logical to add hashing logic to the register endpoint, sure. I don't see it's a problem to leave this logic there :)

Elena-Bruyako (approved):

PR #861, opened 2024-08-12
hibernate-user-service-hw-solution

VArabchuk:


src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,59 @@
...
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
Elena-Bruyako, line 13: Let's make the email unique. We need to check the email during user registration and make the email a unique field in the database.

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,59 @@
+package mate.academy.model;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+
+@Entity
```
Elena-Bruyako, line 8: Add Table annotation

sarakhmen (approved):
lgtm!

PR #866, opened 2024-08-27
hibernate-user-service-hw task solution

SvNadorozhnyi:


src/main/java/mate/academy/Main.java
```diff
@@ -44,7 +50,8 @@ public static void main(String[] args) {
         yesterdayMovieSession.setMovie(fastAndFurious);
         yesterdayMovieSession.setShowTime(LocalDateTime.now().minusDays(1L));
 
-        MovieSessionService movieSessionService = null;
+        MovieSessionService movieSessionService = (MovieSessionService) injector
+                .getInstance(MovieSessionService.class);
```
Elena-Bruyako, line 54: In the mate/academy/Main.main() method create an instance of AuthenticationService using injector and test all methods from it.

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,61 @@
...
+import jakarta.persistence.Table;
+import java.util.Arrays;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String email;
```
Elena-Bruyako, line 16: Let's make the email unique. We need to check the email during user registration and make the email a unique field in the database.

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Inject;
+import mate.academy.model.User;
+import mate.academy.service.AuthenticationService;
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+public class AuthenticationServiceImpl implements AuthenticationService {
```
Elena-Bruyako, line 12:
```suggestion
@Service
public class AuthenticationServiceImpl implements AuthenticationService {
```

Elena-Bruyako (changes requested):

sarakhmen (approved):

PR #874, opened 2024-09-14
Hibernate user service

4Vitalii5:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,51 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            CriteriaBuilder criteriaBuilder = session.getCriteriaBuilder();
```
Elena-Bruyako, line 42: don't use CriteriaBuilder here

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "users")
+public class User {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private String login;
```
Elena-Bruyako, line 15: check common mistakes file, make email (login) unique

Elena-Bruyako (changes requested):

src/main/java/mate/academy/model/User.java
4Vitalii5: Thank you. Added an annotation to make the column unique, but then removed it because I thought checking in the method registry would be enough.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,34 @@
...
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(password.getBytes());
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Can't create hash using SHA-512 algorithm", e);
```
okuzan, line 30: use your constant

src/main/java/mate/academy/model/User.java
okuzan: double check is a common practice :)

okuzan (approved):

PR #878, opened 2024-10-05
First solution

Wojtas1996PL:


src/main/java/mate/academy/Main.java
```diff
@@ -51,5 +58,21 @@ public static void main(String[] args) {
...
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Could not register user", e);
+        }
+
+        try {
+            authenticationService.login("bob.marley@gmail.com", "password");
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("Could not login user", e);
+        }
+
```
MateuszRostek, line 76: What about when we trying to register when email address is taken? Please add this case.

src/main/java/mate/academy/Main.java
```diff
@@ -2,14 +2,21 @@
...
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.impl.AuthenticationServiceImpl;
 
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.academy");
+
     public static void main(String[] args) {
         MovieService movieService = null;
```
MateuszRostek, line 21: You should also initialize Services by using the injector.

src/main/java/mate/academy/model/User.java
```diff
@@ -0,0 +1,38 @@
+package mate.academy.model;
+
+public class User {
```
MateuszRostek, line 3: This is an entity, so please add required annotations and field (id).

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+            return user;
+        } else {
+            throw new AuthenticationException("Password is wrong for user with email " + email);
+        }
+    }
+
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (!userFromDbOptional.isEmpty()) {
```
MateuszRostek, line 36:
```suggestion
        if (userFromDbOptional.isPresent()) {
```

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,31 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class HashUtil {
+    private static final String CRYPTO_ALGORITHM = "SHA-512";
+
+    public static byte[] getSalt() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] salt = new byte[16];
```
MateuszRostek, line 12: Don't use magic numbers, make it a constant.

MateuszRostek (changes requested):
Nice try! See some comments and implement them. Also, you should add the `hibernate.cfg.xml` file.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,32 @@
...
+        secureRandom.nextBytes(salt);
+        return salt;
+    }
+
+    public static String hashPassword(String password, byte[] salt) {
+        StringBuilder hashedPassword = new StringBuilder();
+        try {
+            MessageDigest messageDigest = MessageDigest.getInstance(CRYPTO_ALGORITHM);
+            messageDigest.update(salt);
+            byte[] digest = messageDigest.digest(salt);
```
slade13, line 23: You're digesting the salt, not the password. You should pass the password to the 'digest' method of the 'MessageDigest' instance.

src/main/java/mate/academy/util/HashUtil.java
```diff
@@ -0,0 +1,32 @@
...
+            for (byte b : digest) {
+                hashedPassword.append(String.format("%02x", b));
+            }
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("Could not create hash using SHA-512 algorithm");
```
slade13, line 28: The exception message 'Could not create hash using SHA-512 algorithm' is misleading. This exception is thrown when the specified algorithm is not available, not when there's a problem creating the hash. Consider changing the message to 'SHA-512 algorithm not available.'

src/main/java/mate/academy/service/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public User register(String email, String password) throws RegistrationException {
+        Optional<User> userFromDbOptional = userService.findByEmail(email);
+        if (userFromDbOptional.isPresent()) {
+            throw new RegistrationException("User with email " + email + " already exist");
+        }
+        User user = new User();
+        user.setEmail(email);
+        user.setPassword(password);
+        return userService.add(user);
```
slade13, line 42: The creation of the user in the database should be a transactional operation to avoid data inconsistency in case of an error.

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +56,35 @@ public static void main(String[] args) {
...
 
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
                         fastAndFurious.getId(), LocalDate.now()));
+
+        AuthenticationService authenticationService = (AuthenticationService) injector
+                .getInstance(AuthenticationServiceImpl.class);
+
+        try {
+            authenticationService.register("bob.marley@gmail.com", "password");
```
slade13, line 72: Please handle the RegistrationException and AuthenticationException more gracefully. Throwing a RuntimeException makes it harder to track the actual place where the original exception was thrown. Also, it's better practice to log the exception instead of re-throwing it.

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Inject
+    private UserDao userDao;
+
+    @Override
+    public User add(User user) {
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
```
slade13, line 19: Hashing the password should be the responsibility of the AuthenticationService during registration, not the UserService.

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import mate.academy.service.UserService;
+import mate.academy.util.HashUtil;
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Inject
+    private UserDao userDao;
+
+    @Override
+    public User add(User user) {
```
slade13, line 17: Consider validating the user object before adding it to the database.

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+        user.setSalt(HashUtil.getSalt());
+        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
+        return userDao.add(user);
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
```
slade13, line 24: Consider validating the email before using it to find a user.

src/main/java/mate/academy/service/impl/UserServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import mate.academy.util.HashUtil;
+
+@Service
+public class UserServiceImpl implements UserService {
+    @Inject
+    private UserDao userDao;
+
+    @Override
+    public User add(User user) {
+        user.setSalt(HashUtil.getSalt());
```
slade13, line 18: Generating and setting the salt should be the responsibility of the AuthenticationService during registration, not the UserService.

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public Optional<User> findByEmail(String email) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from user u where email = :email", User.class)
```
slade13, line 39: The 'from user' in the query should be 'from User' as HQL queries are case-sensitive and the entity names should start with a capital letter.

src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.persist(user);
+            transaction.commit();
+            return user;
+        } catch (Exception e) {
+            if (transaction != null) {
```
slade13, line 25: You should check transaction status before rollback. If transaction is already committed, rolled back or never opened TransactionException would be thrown.

slade13 (approved):
I'm approving your changes but please fix a bug with password hashing in: src/main/java/mate/academy/util/HashUtil.java

PR #949, opened 2025-05-09
added with errors v1.0

kostya-savchenko:


kostya-savchenko:
Password Security: In your AuthenticationServiceImpl, the register method currently stores passwords in plain text. For security and to meet the requirements, you must generate a salt, hash the password with it, and store both the hashed password and the salt in the User object before saving. This is essential for secure authentication and is a critical issue.
    
But I have generation salt in the UserServiceImpl 

    @Override
    public User save(User user) {
        user.setSalt(HashUtil.getSalt());
        user.setPassword(HashUtil.hashPassword(user.getPassword(), user.getSalt()));
        return userDao.save(user);
    }

PR #1007, opened 2026-03-20
Logic Implemented

qreqit:


src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -0,0 +1,19 @@
+package mate.academy.exception;
+
+public class AuthenticationException extends RuntimeException {
```
fetis-off, line 3: A am wondering do you really use all of these constructors? Please leave only necessary constructors, which are used in your implementation

fetis-off (approved):
Good work. Please resolve one comment before merging

PR #1012, opened 2026-04-11
first try

Dima8899:


src/main/java/mate/academy/dao/impl/UserDaoImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import org.hibernate.query.Query;
+
+@Dao
+public class UserDaoImpl implements UserDao {
+    private final SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+
+    @Override
+    public User add(User user) {
+        Transaction transaction = null;
+        try (Session session = sessionFactory.openSession()) {
```
Elena-Bruyako, line 21: don't use transaction and try with resources, if exception happen session will close before catch block

use try catch finally

Elena-Bruyako (changes requested):
see comment

